PR #95, opened 2020-07-02
slashing/termination from failed windowed post + fast retrieval benchmarks

### LINKED ISSUE #12: [test request] verify recovery from failed windowed PoST
**What would you like us to test?**

Verify that miners can recover from faults in window PoST.

**Technical implementation details.**

Except from call (@yusefnapora's notes):

```
- lukasz:
    - in fil, you prove each sector once per day, which we currently do in 36 windows per day (every 40 mins)
    - each sector is assigned to a proving window, miner computes window PoST and submits it
    - if proof is missing, sector is marked faulty - miner gets 2 week window to recover the sector
      - recovery process:
        - before the window is open, you can declare that you will recover the faulty sector, then you have to submit the proof in the correct window.
    - raul q: where does slashing come into this process?
       - lukasz: when your sector is reported faulty, the power for that sector is removed
       - also, when your sector is marked faulty, you pay a fee (b/c the storage is missing)
    - actors are responsible for proving that you didn't prove a sector & tracking what was lost/recovered
    - lotus needs to look at what sectors it needs to prove in the next window
      - also needs to detect faults on disk
    - proving works using a windowed PoST scheduler
      - looks at what blocks come in from the network
        - for every block, asks the miner actor if there's a new proving window open
          - if so, it waits a few blocks (for confidience threshold), computes a window PoST
        - if there's a chain-reorg, the client needs to detect whether their proof was included in the final chain and resubmit if not
    - before running windowed PoST, we scan sectors to make sure we can read from disk
      - it's still possible that a fault can occur even if the sector is readable during the scan (could be intermittent failure, or the fault exists in a part of the sector that wasn't scanned)
    - this will likely be hard to test due to the long time frames (fault recovery period lasts 2 weeks)
      - even mock tests can take minutes
      - magik: we have some tests that claim block time is 25seconds, but actually produce block in a few milliseconds. Still takes a while, but much faster than with stock params
      - why: time is actually measured in blocks, not wall-clock time
        - in tests, we can say "mine the next block" at arbitrary times, can sync/wait on everyone receiving that block
      - wyatt: while you can fake the block time to an extent, running the proofs takes real time, so it's not possible to completely elimitate time. e.g. if you're testing recovery, computing the power table reqires real proofs / time
      - magik: mining a block with 2k sectors takes 30 seconds
      - raul: are proofs necessary for all tests, or can they be faked?
        - magik: not necessary for many tests. we already have fairly powerful & accurate mocks for proofs
          - they behave in much the same way, but aren't safe / secure.
          - using the DI system, you can easily swap out any component with a few lines of code
            - uber/fx with a custom layer on top
```

**What should we measure?**

- Behaviour under chain reorgs, after posting proof.
- Behaviour under chain reorgs, prior to posting proof.
- Behaviour under network partitions.
- Failed sector reads during proof generation (transitorily corrupt the sector?).
- Slashing effective after 2 week period (how can we accelerate the grace period?)

**Which components are involved?**

TBD.

**On a scale from 0-10, what's the proposed _**discomfort factor**_? In other words, how uncomfortable would you be if we went live **without** having tested this? Explain why.**

TBD.

**Additional remarks.**

Failed window PoST nuked previous testnets.

Requestor: @travisperson.
More technical details: @magik6k, @whyrusleeping.

### LINKED ISSUE #100: [test scenario] windowed PoSt - verify faulted state and slashing
# See #118 instead

---

**Describe the test scenario.**

This is part of the windowed PoSt epic. Verify that if a miner fails to post a sector proof in their window, the sector is declared as faulted by the network and [Sector Penalties](https://filecoin-project.github.io/specs/#storage-fault-slashing) are levied upon the miner.

```
1. Fault Fee (FF): A penalty that a miner incurs for each day a miner’s sector is offline (either declared or because it was detected by the network).
2. Sector Penalty (SP): A penalty that a miner incurs for a faulted sector that was not declared faulted before a WindowPoSt check occurs.
a) The sector will pay FF after incurring an SP when the fault is detected.
3. Termination Fee (TP): A penalty that a miner incurs when a sector is voluntarily or involuntarily terminated and is removed from the network.
```

Related code as: `specs-actors/actors/builtin/miner/monies.go`

**What should we measure?**

1. Epochs elapsed until the sector is marked as faulty by the network.
2. Penalties incurred - FF, SP and TP - and when they are incurred.

**Discomfort factor (0-10).**

High (8-10).

**Additional remarks.**

Read https://filecoin-project.github.io/specs/#storage-fault-slashing.

### LINKED ISSUE #111: [test scenario] slashing - verify fault fee applied when sector is faulted
**Describe the test scenario.**

This issue is part of the slashing epic (see https://github.com/filecoin-project/oni/issues/21) and covers the basic application of the Fault Fee to a miner that has a faulted sector.

The high-level plan is to engineer a fault and wait for the proving window to complete, then verify that the miner gets slashed by the expected amount.

**Provide any background and technical implementation details.**

The [slashing google doc](https://docs.google.com/document/d/1m9NU-NfaFqg8SdwMm9hf-1wzPHnFC-9PJSub8ho9R2E/edit#) has more info about definitions.

The Fault Fee should be applied whether the miner declares a fault or not, although if the miner doesn't declare the fault, they should receive an additional Sector Penalty.

During the fault, the miner is also not eligible for any block rewards for the faulted sectors.

We'll almost certainly want to run with reduced window periods using @nonsense's spec-actors PR branch: https://github.com/filecoin-project/specs-actors/pull/595 - otherwise we'll have to actually run for a day before the slashing gets applied.

**To figure out:**

- [ ] how to trigger a storage fault without taking the whole miner node offline
  - with fake proof verification, a "valid" proof is just the bytes `valid proof` in utf-8: https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L865
  - the fake proof is [generated in `StorageWpp.ComputeProof`](https://github.com/filecoin-project/lotus/blob/fe4efa0e0ee045d38f63f4955d01cbf3e36bc41f/storage/miner.go#L178) if `build.InsecurePoStValidation` is true, otherwise it calls through to `GenerateWinningPoSt` on a `storage.Prover`.
    - this means we can't easily inject a fake `Prover` that fails on command while `build.InsecurePoStValidation` is set, since our `Prover` won't get used if the flag is true. If we set the flag to false, it will use our custom `Prover`, but will also use real proof verification, which will be slower. But maybe this test should use real proofs anyway for greater fidelity.
   - instead of generating an invalid proof, maybe we can hook into the code that submits the proof, and just prevent the proof from being sent out.
  - ideally, we want to be able to control whether the miner declares the sector as faulty or tries to "get away with it" and risk the Sector Penalty

- [ ] How to calculate the expected slash amount. The doc refers to a constant that gets multiplied by `BR(t)` which is the naively approximated block reward over the proving window, starting from epoch `t`. 
  - I'm assuming `t` is the start of the proving window - should verify
  - the doc refers to "the next 24 hours" for calculating `BR(t)` - i'm assuming this _really_ means "over the proving window", but if it's really set to 24 hours and we're changing the proving window, we may get slashed by a much greater amount than we expect.
  - where in the code is this calculated? probably specs-actors but I haven't yet verified.

- [ ] How to verify that the slash took place
  - There's a [`GetMinerSlashed` method in `stmgr/utils.go`](https://github.com/filecoin-project/lotus/blob/516e31d37cd70e38c8d0529d1fe3f774a0d5ca6d/chain/stmgr/utils.go#L274) that may be useful for verifying whether a miner was slashed. It doesn't return the amount though, just a `bool`.
  - There are also `GetMinerFaults` and `GetMinerRecoveries` methods in the same file that may be useful.
  - We also should verify that the miner didn't receive any block reward for the faulted sector.

**What should we measure?**

Whether the miner was slashed and by how much.

**Discomfort factor (0-10).**

 

**Additional remarks.**

### LINKED ISSUE #113: [test scenario] slashing - verify faulted sector recovery (re-proving)
**Describe the test scenario.**

After a miner has been slashed for failing to prove a sector during a proving window, they can recover the sector by re-proving it at the next window (up to 14 times before the sector is marked as terminated).

We want a test case that will verify that a miner can recover a faulty sector, and is no longer penalized once the sector is marked as recovered.

**Provide any background and technical implementation details.**

Assuming that we've already done #111 and figured out how to trigger a fault on command, we should be able to just flip the switch after the initial fault and wait for the next proving window. The miner should submit a valid proof and the sector should be marked as valid.

We should also test the scenario where we _claim_ that the sector has been recovered, but don't actually submit the proof during the next window. According to the doc, this should trigger a Sector Penalty, in addition to the Fault Fee for missing the proving window.

According to the [slashing doc](https://docs.google.com/document/d/1m9NU-NfaFqg8SdwMm9hf-1wzPHnFC-9PJSub8ho9R2E/edit#):

> To recover any sectors from the faulted state, a miner needs to lock additional funds such that the total LockedFunds is at least the sum of all initial pledges for all active sectors. In other words, StorageMinerActor needs to keep track of an individual sector’s Initial Pledge for TF computation and a total InitialPledge for all active sectors.

So we could potentially try to prove that it's impossible to recover a sector if you don't have enough Fil to lock up the additional collateral.

**What should we measure?**

In the case of a successful recovery, we should verify that the miner is not slashed for any proving windows after the sector is recovered. We should also verify that they received block rewards for that sector if they were eligible for rewards.

**Discomfort factor (0-10).**

 

**Additional remarks.**

### LINKED ISSUE #120: [test scenario] slashing - verify sector termination fee is levied upon 14 consecutive missed proving windows
**Describe the test scenario.**

Part of slashing epic (#21) - see [slashing google doc](https://docs.google.com/document/d/1m9NU-NfaFqg8SdwMm9hf-1wzPHnFC-9PJSub8ho9R2E/edit#) for more info.

If a sector is faulted for 14 consecutive proving windows, or if a miner declares the sector to be terminated, they get hit with a Termination Fee (in addition to the Fault Fees and whatever Sector Penalties they might have accrued during the fault).

**Provide any background and technical implementation details.**

We should test both ways of reaching the terminal state:

- wait 14 proving windows and check that the sector was marked as terminated and the fee was applied
- have the miner declare the sector terminated

**What should we measure?**

- verify that the sector was marked as terminated in chain state
- verify that the miner was slashed by the termination fee amount (plus other penalties that stack up during the fault)

**Discomfort factor (0-10).**

 

**Additional remarks.**

nonsense:
This PR is the initial test cases for:
1. testing slashing and recovery scenarios. (Fault Fee (FF), Sector Penalty (SP), Termination Fee (TP))
2. fast retrieval benchmarks

As part of outputs this PR also:
1. Generates chain `diff` on every height (this only makes sense if we don't have reorgs)
2. Generates chain HTML state on every height
3. Updated `RetrieveData` signature, so that it doesn't panic, but returns an error (the client that makes a deal with the miner that goes offline is expected to get an error when trying to retrieve its data).

---

FOLLOW UP TODO:
- [ ] 1. automatic detection of full slashing (incl. termination fee) is not automated, as slashing changed recently and we can no longer expect just 3 slashing conditions; needs to be improved once we gain understanding when watching slashing happen in manual runs;
- [ ] 2. automatic detection of partial slashing (miner getting FF and SP, but not TP) is not automated, as miners do not recover correctly after being slashed - it is not clear if this is a bug on Lotus / specs-actors or in the test, just yet. Reported to Lotus team for now.
- [ ] 3. signaling for 1 and 2 in the test plan is not perfect. The idea is to have 2 healthy miners, 1 miner to be parially slashed, and 1 miner to be fully slashed, where the 2 healthy miners receive signals and confirm that a given miner is partially slashed and a given miner is fully slashed.
- [ ] 4. consolidate `FetchChainState`, `UpdateChainState` and `tstats.Collect` - they all have to do with capturing state of the blockchain from the perspective on a given miner on a given tipset height.
a) `FetchChainState` - generates HTML trace of messages for every epoch
b) `UpdateChainState` - generates `chain-state.ndjson` for consumption in Jupyter and `diff` between `height-1 .. height` to detect slashing conditions.
c) `tstats.Collect` - pushes data points to InfluxDB for the dashboard of the blockchain in Grafana.

---

BLOCKERS / PROBLEMS / OPEN QUESTIONS:

- [ ] 1. Races and panics on shutdown: [lotus-panic.txt](https://github.com/filecoin-project/oni/files/4972272/lotus-panic.txt) ; https://github.com/filecoin-project/lotus/issues/2572
- [ ] 2. Miners do not recover after being stopped and started, they do not regain their power, and do not mine blocks, so it is hard to test `recovery`.
- [ ] 3. Slash conditions changed very recently (1-2 days ago) so we are not detecting changes in balances right now to prove slashings… in 2 weeks from now maybe we can do that through the journal and detect an event, rather than calculate balances difference.

---

Fixes #100.
Fixes #12.
Fixes #113.
Fixes #111.
Fixes #120.

^^ There's more work to do, but I'm happy closing those issues for now and specifying the follow-ups with more detail.

Addresses #157.


lotus-soup/testkit/role_miner.go
```diff
@@ -394,6 +403,8 @@ func (m *LotusMiner) RunDefault() error {
 		return err
 	}
 
+	time.Sleep(3600 * time.Second)
```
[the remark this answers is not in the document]
nonsense: This test is not yet ready. I don't want my instances to exit, while I am inspecting what happens in the network due to the `biserk` miner. So I sleep for 1 hour, and manually kill the test.

docker-images/Dockerfile.oni-runtime-debug
```diff
@@ -13,7 +13,7 @@ COPY /proof-parameters.json /
 RUN paramfetch 2048 /proof-parameters.json
 
 ## for debug purposes
-RUN apt update && apt install -y mesa-opencl-icd ocl-icd-opencl-dev gcc git bzr jq pkg-config curl && git clone https://github.com/filecoin-project/lotus.git && cd lotus/ && git checkout next && make clean && make all && make install
+RUN apt update && apt install -y mesa-opencl-icd ocl-icd-opencl-dev gcc git bzr jq pkg-config curl && git clone https://github.com/filecoin-project/lotus.git && cd lotus/ && git checkout 7c73dbf && make clean && make all && make install
```
yusefnapora, line 16: maybe make the git ref a build variable?

lotus-soup/rfwp/chain_state.go
```diff
@@ -456,7 +456,7 @@ func provingDeadlines(t *testkit.TestEnvironment, m *testkit.LotusMiner, maddr a
 		}
 	}
 
-	infos := make([]DeadlineInfo, len(deadlines.Due))
+	infos := make([]DeadlineInfo, 0, len(deadlines.Due))
```
nonsense, line 459: @yusefnapora FYI

nonsense:
@raulk @yusefnapora I suggest we review and merge this and continue iterating on the recovery test (already have another draft PR on top of this), as long as it is not interfering with other work. otherwise we keep on hitting merge conflicts, due to upgrades, and what not.

docker-images/Dockerfile.oni-runtime-debug
```diff
@@ -12,8 +12,10 @@ RUN go get github.com/filecoin-project/go-paramfetch/paramfetch
 COPY /proof-parameters.json /
 RUN paramfetch 2048 /proof-parameters.json
 
+ARG LOTUS_COMMIT=e21ea53
```
raulk, line 15: I think we might be better served setting this value as an argument in the `Makefile`.

lotus-soup/deals_e2e.go
```diff
@@ -108,3 +127,50 @@ func dealsE2E(t *testkit.TestEnvironment) error {
...
+
+// filToAttoFil converts a fractional filecoin value into AttoFIL, rounding if necessary
+func filToAttoFil(f float64) big.Int {
+	a := mbig.NewFloat(f)
+	a.Mul(a, mbig.NewFloat(float64(build.FilecoinPrecision)))
+	i, _ := a.Int(nil)
+	return big.Int{Int: i}
+}
+
+func initPaymentChannel(t *testkit.TestEnvironment, ctx context.Context, cl *testkit.LotusClient, minerAddr testkit.MinerAddressesMsg) error {
```
raulk, line 139: Can we move this to the `testkit` library under a `paych.go` file, and make it public?
Then we can dedup this logic in the `paych_stress` and be DRYer.

lotus-soup/init.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 func init() {
-	build.BlockDelaySecs = 2
+	build.BlockDelaySecs = 1
```
raulk, line 18: Is this intentional?

lotus-soup/rfwp/diffs.go
```diff
@@ -0,0 +1,295 @@
...
+			fmt.Fprintf(w, "%s diff of             |\n", toCharStr(i))
+		}
+
+		for difference, heights := range cs.DiffCmp[maddr][valueName] {
+			fmt.Fprintf(w, "%s diff of %30v at heights %v\n", toCharStr(i), difference, heights)
+		}
+	}
+}
+
+func recordDiff(mi *MinerInfo, ps *ProvingInfoState, height abi.ChainEpoch) {
```
raulk, line 75: I think we might end up reusing some of this diff logic for our test-vector-driven testing.

lotus-soup/rfwp/e2e.go
```diff
@@ -0,0 +1,347 @@
...
+	go func() {
+		errc <- eg.Wait()
+	}()
+
+	select {
+	case err := <-errc:
+		if err != nil {
+			return err
+		}
+	case err := <-t.SyncClient.MustBarrier(ctx, testkit.StateAbortTest, 1).C:
```
raulk, line 98: This is nice; we should propagate this pattern to the SDK.

lotus-soup/init.go
nonsense: Not really, just wanted the test to be faster. Will revert. (I didn't see any difference between 2 and 1.)

PR #117, opened 2020-07-08
consistent compositions

nonsense:
1. Make sure all our compositions use the build tag `testground`, otherwise they won't compile.
2. Make sure all our compositions use the proper exposed_ports format.

### DIFF
--- a/lotus-soup/compositions/composition-drand-halt.toml
+++ b/lotus-soup/compositions/composition-drand-halt.toml
@@ -12,11 +12,11 @@
 [global.build]
   selectors = ["testground"]
 
-[global.build_config]
-  enable_go_build_cache = true
-
 [global.run_config]
   exposed_ports = { pprof = "6060", node_rpc = "1234", miner_rpc = "2345" }
+
+[global.build_config]
+  enable_go_build_cache = true
 
 [global.run.test_params]
   clients = "1"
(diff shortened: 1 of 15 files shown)


lotus-soup/compositions/composition-drand-halt.toml
```diff
@@ -12,12 +12,12 @@
...
 
-[global.build_config]
-  enable_go_build_cache = true
-
 [global.run_config]
   exposed_ports = { pprof = "6060", node_rpc = "1234", miner_rpc = "2345" }
 
+[global.build_config]
+  enable_go_build_cache = true
+
```
raulk, line 20: uber nit: I usually like to specify build/run in chronological order: build first, run second.

lotus-soup/compositions/composition-local-drand.toml
```diff
@@ -9,6 +9,12 @@
   builder = "docker:go"
   runner = "local:docker"
 
+[global.build]
+  selectors = ["testground"]
+
+[global.run_config]
+  exposed_ports = { pprof = "6060", node_rpc = "1234", miner_rpc = "2345" }
```
raulk, line 16: We should really have ports inherited from the manifest. We should be able to just add them once in the manifest...

lotus-soup/compositions/composition-drand-halt.toml
nonsense: I changed it so that all compositions are consistent. and the build_config changes between runners.

lotus-soup/compositions/composition-local-drand.toml
nonsense: Same applies for the `selector` I'd say.

PR #122, opened 2020-07-08
print auth token ; set seal delay ; disable permissioned api proxies ; bump balances

nonsense:
1. Set `seal delay` to 30sec. By default it is 1 hour. 30sec. should be fine with our block delay of 2 sec. and propagation delay of 1sec. This will be added once `next` is merged to `master`, so we need also an update to `go.mod`.
2. Print out the auth token, so that we can use it to have access to all internal APIs.
3. Disable API permissioned proxy.
4. Adding a `debug` runtime image, that bundles `lll` and `lsm` so that we can quickly query debug info on containers.
5. Bump balances, as 2000 results in errors when generating genesis block.
6. Update `deals e2e` and `deals stress` tests due to changed APIs.

---

Follow up TODO:
- [ ] move HISTORY.md to README.md


lotus-soup/testkit/role_miner.go
```diff
@@ -245,6 +250,20 @@ func PrepareMiner(t *TestEnvironment) (*LotusMiner, error) {
...
+		return nil, err
+	}
+
+	// print out the admin auth token
+	token, err := n.MinerApi.AuthNew(ctx, apistruct.AllPermissions)
+	if err != nil {
+		return nil, err
+	}
+
+	t.RecordMessage("Auth token: %s", string(token))
```
nonsense, line 265: Now that we no longer start permissioned APIs, this is not 100% necessary, but probably worth to have this token generated and printed out just in case.

docker-images/Dockerfile.oni-runtime-debug
```diff
@@ -0,0 +1,28 @@
+ARG GO_VERSION=1.14.2
```
raulk, line 1: Now that we have more images, can we turn the `HISTORY.md` into a `README.md` (moving all its content to a section `## Image version history`, and describing in one line what each of the three images contains/does?

lotus-soup/compositions/composition-k8s.toml
```diff
@@ -25,7 +25,7 @@
   clients = "1"
   miners = "1"
   genesis_timestamp_offset = "100000"
-  balance = "2000"
+  balance = "20000000"
```
raulk, line 28: These balances will work for maximum 100 nodes, as TotalFilecoin is 2B:

https://github.com/filecoin-project/lotus/blob/3b826e41f6afbcccd2f93c7f3fcd89f1574b4f84/build/params_shared_vals.go#L64

Mind putting a comment here and in all other compositions? Pretty sure if we go above 100 nodes, it'll be a head scratcher and we'll end up burning time debugging why things break.

lotus-soup/go.mod
```diff
@@ -7,11 +7,11 @@ require (
...
 	github.com/drand/drand v0.9.2-0.20200616080806-a94e9c1636a4
 	github.com/filecoin-project/go-address v0.0.2-0.20200504173055-8b6f2fb2b3ef
-	github.com/filecoin-project/go-fil-markets v0.3.2-0.20200706104419-7c180fe156d4
+	github.com/filecoin-project/go-fil-markets v0.4.0
 	github.com/filecoin-project/go-jsonrpc v0.1.1-0.20200602181149-522144ab4e24
 	github.com/filecoin-project/go-storedcounter v0.0.0-20200421200003-1c99c62e8a5b
-	github.com/filecoin-project/lotus v0.4.2-0.20200706172415-cf6ac44b6ec5
-	github.com/filecoin-project/specs-actors v0.6.2-0.20200702170846-2cd72643a5cf
+	github.com/filecoin-project/lotus v0.4.2-0.20200708212122-7d841dbfa81c
+	github.com/filecoin-project/specs-actors v0.7.1
```
raulk, line 14: So we're switching to `next` entirely on master, I guess?

nonsense: Yes!

PR #123, opened 2020-07-08
disable permissioned proxy /2

nonsense:
This PR is disabling the permissioned proxy for the miner and full node APIs.


docker-images/Dockerfile.oni-runtime-debug
```diff
@@ -0,0 +1,28 @@
...
+
+RUN apt-get update && apt-get install -y ca-certificates llvm clang mesa-opencl-icd ocl-icd-opencl-dev jq gcc pkg-config net-tools netcat traceroute iputils-ping wget vim curl telnet iproute2 dnsutils
+COPY --from=downloader /var/tmp/filecoin-proof-parameters /var/tmp/filecoin-proof-parameters
+
+## for debug purposes
+COPY --from=downloader /usr/local/bin/lotus /usr/local/bin/lll
+COPY --from=downloader /usr/local/bin/lotus-storage-miner /usr/local/bin/lsm
+
+ENV FULLNODE_API_INFO="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.UYhed-XqqHEfSBZ1_hrwPvB3VBX3N06dPCGcZqYXObM:/ip4/127.0.0.1/tcp/1234/http"
+ENV STORAGE_API_INFO="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.UYhed-XqqHEfSBZ1_hrwPvB3VBX3N06dPCGcZqYXObM:/ip4/127.0.0.1/tcp/2345/http"
```
nonsense, line 28: I am open to other idea on how to solve this, but since overloading `Verify` didn't work and we still need a token, I decided to just make it static.

---

Obviously without this the flow I go through so far is manually check logs, extract the token and set the env var in the shell when inspecting values with `lsm` and `lll`.

docker-images/Dockerfile.oni-runtime-debug
```diff
@@ -0,0 +1,28 @@
...
+## 2. Then copy over the parameters (which could change).
+## 3. Trigger the download.
+## Output will be in /var/tmp/filecoin-proof-parameters.
+
+RUN go get github.com/filecoin-project/go-paramfetch/paramfetch
+COPY /proof-parameters.json /
+RUN paramfetch 2048 /proof-parameters.json
+
+## for debug purposes
+RUN apt update && apt install -y mesa-opencl-icd ocl-icd-opencl-dev gcc git bzr jq pkg-config curl && git clone https://github.com/filecoin-project/lotus.git && cd lotus/ && git checkout next && make clean && make all && make install
```
nonsense, line 16: Depending on how often these tools (`lotus` and `lotus-storage-miner`) change, we might want to move these utilities directly in the `testplan` Dockerfile. For now I have had some success by bundling them with the `runtime` image, but it is rather annoying when they change, as I have to recompile, tag, push images.

Issue #125, opened 2020-07-08
Notes on how to run the tests


schomatis:
Something that describes how can I run one of the tests in this repo from scratch.

raulk:
@schomatis: @nonsense took care of this. Mind going through the [instructions on the README](https://github.com/filecoin-project/oni#running-the-test-cases) and trying them out?

PR #129, opened 2020-07-09
improve readme with instructions on how to run test cases

nonsense:
Fixes https://github.com/filecoin-project/oni/issues/125

---

TODO:
- [ ] document and refactor our compositions files and describe what each composition is running or at least attempting to test. I think it doesn't make sense to write in prose in details all the values for the composition, but at least mention the groups being started and what the test does.


nonsense:
I consider the TODO as out of scope for this PR, but def something we should do sooner rather than later.

PR #121, opened 2020-07-08
wip - basic chain halting / recovery

yusefnapora:

PR #134, opened 2020-07-09
implement Halt/Resume for miner nodes

yusefnapora:
This pulls out the suspend / resume code from #121 and removes the test plan skeleton I started for the chain recovery scenario.

Now you can add a `suspend_events` param to miner nodes, although currently with synchronized mining you have to halt / resume all miners at once. If you only halt one, the rest will still be waiting on the sync barrier.

### DIFF
--- a/lotus-soup/testkit/role_miner.go
+++ b/lotus-soup/testkit/role_miner.go
@@ -17,6 +17,7 @@ import (
 	"github.com/filecoin-project/lotus/api/apistruct"
 	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/chain/actors"
+	"github.com/filecoin-project/lotus/chain/gen"
 	genesis_chain "github.com/filecoin-project/lotus/chain/gen/genesis"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/wallet"
@@ -25,6 +26,7 @@ import (
 	"github.com/filecoin-project/lotus/node"
 	"github.com/filecoin-project/lotus/node/impl"
 	"github.com/filecoin-project/lotus/node/modules"
+	"github.com/filecoin-project/lotus/node/modules/dtypes"
 	"github.com/filecoin-project/lotus/node/repo"
 	"github.com/filecoin-project/specs-actors/actors/abi"
 	"github.com/filecoin-project/specs-actors/actors/builtin"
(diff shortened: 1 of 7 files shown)


nonsense:
Overall LGTM in terms of code, but naming is getting a bit off, so I am adding a few suggestion for improvements - could be part of this PR or follow up PRs within the team.

lotus-soup/testkit/role_miner.go
```diff
@@ -114,7 +121,13 @@ func PrepareMiner(t *TestEnvironment) (*LotusMiner, error) {
 		return nil, err
 	}
 
-	// prepare the repo
+	// create the node
+	// we need both a full node _and_ and storage miner node
+	n := &LotusNode{}
+
+	m := &LotusMiner{LotusNode: n, t: t}
```
nonsense, line 128: We are overloading the term `miner`. I suggest we rename `LotusMiner` to `MinerNode`, to be explicit that this is not just the subsystem that produces blocks, but the full libp2p process/node.

lotus-soup/testkit/role_miner.go
```diff
@@ -114,7 +121,13 @@ func PrepareMiner(t *TestEnvironment) (*LotusMiner, error) {
 		return nil, err
 	}
 
-	// prepare the repo
+	// create the node
+	// we need both a full node _and_ and storage miner node
+	n := &LotusNode{}
```
nonsense, line 126: Similar to the comment below, I suggest we rename this to `FullNode`.

lotus-soup/testkit/role_miner.go
```diff
@@ -45,6 +50,8 @@ type LotusMiner struct {
 	*LotusNode
 
 	t *TestEnvironment
+
+	minerInstance *miner.Miner
```
nonsense, line 54: I think this could be called just `miner`. However it is important to note that this is just the `block producing` subsystem, and not the `MinerNode` - so I suggest a few renames further down.

lotus-soup/testkit/role_miner.go
```diff
@@ -432,7 +462,24 @@ func (m *LotusMiner) RunDefault() error {
 	return nil
 }
 
-func startStorageMinerAPIServer(t *TestEnvironment, repo *repo.MemRepo, minerApi api.StorageMiner) error {
+func (m *LotusMiner) Halt() {
+	m.t.RecordMessage("halting miner")
```
nonsense, line 466: I understand this as if we are stopping the whole miner node, whereas it is just stopping the block producing part of it. I think it makes sense being explicit about it.

raulk:
@yusefnapora we should rebase this on master after merging #95, which introduces the ability to start/stop the miner node entirely (not just the mining subsystem). Can we rewire the suspension state machine to those APIs?

PR #143, opened 2020-07-13
WIP proper synchronized mining 

raulk:


lotus-soup/testkit/role_miner.go
```diff
@@ -202,10 +218,17 @@ func PrepareMiner(t *TestEnvironment) (*LotusMiner, error) {
...
 		mineBlock := make(chan func(bool, error))
 		minerOpts = append(minerOpts,
-			node.Override(new(*miner.Miner), miner.NewTestMiner(mineBlock, minerAddr)))
+			node.Override(new(*miner.Miner), miner.NewTestMiner(mineBlock, minerAddr)),
+		)
+
+		// replace the clock, setting it to the genesis timestamp + 1 epoch
+		// we are now ready to mine (and receive) the first block.
+		m.MockClock = clock.NewMock()
+		m.MockClock.Set(time.Unix(int64(genesisMsg.TimestampUnix+build.BlockDelaySecs), 0))
```
schomatis, line 230: Since we increase the clock upfront at the beginning of the `for` in `launchMinerCoordination()` consider not adding this initial `BlockDelaySecs` here (the miners complain they are mining in the past and receiving blocks with large delays). Still, we could consider a very slight time shift here to avoid the edge case of comparing _exactly_ equal times (which many parts of the Lotus code might not be prepared for).

raulk: Yep, good recommendation. I fixed this in my overhaul, which I just pushed to this branch.

lotus-soup/testkit/node_time.go
```diff
@@ -0,0 +1,98 @@
...
+// local signals to advance the clock to the next epoch. Such signals should
+// correspond to some local event that implies readiness to advance, such as
+// "mining round completed" events from synchronized mining.
+//
+// Upon receiving a local advance signal, it forwards the clock by build.BlockDelaySecs,
+// thus progressing one chain epoch. It then signals this fact to all other
+// nodes participating in the test via the synchronization service.
+//
+// Once all miners and clients have advanced, this function releases the new
+// abi.ChainEpoch on the returned channel, thus signalling that all nodes are
```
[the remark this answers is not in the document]
raulk: Thanks for pointing this out. Corrected in https://github.com/filecoin-project/oni/pull/143/commits/193d78fbd4ded7fe4fd343782d6ed3ecf6697e1d.

raulk:
@yusefnapora https://github.com/filecoin-project/lotus/pull/2362/commits/55d88440eb5cbbd31950bdf664b50b5456372ede

raulk:
@yusefnapora 

```
Jul 14 14:52:33.312636  INFO    101.6294s      ERROR << instance   4 >> 2020-07-14T10:52:33.312-0400    WARN    sub     sub/incoming.go:81      Received block with large delay 25 from miner t01001    {"req_id": "8ac15ff3"}
```

is a bit of a red herring. The [block validator](https://github.com/filecoin-project/lotus/blob/13de81b3b27926fd5db1e7334d540f8e6d72c774/chain/sub/incoming.go#L143) is not discarding blocks we've already received, so it appears that gossipsub delivers can deliver the same block multiple times. The first time we receive the block, we incorporate it to our HEAD, and when all blocks are received, we advance time. If due to gossipsub amplification, a duplicate block is received after we've fast-forwarded the clock 25s, it will still be delivered and will cause this warning.

schomatis:
> The block validator is not discarding blocks we've already received,

Not totally familiar with gossipsub internals but we should drop (valid) repeated blocks in 

https://github.com/filecoin-project/lotus/blob/13de81b3b27926fd5db1e7334d540f8e6d72c774/chain/sub/incoming.go#L220-L225

lotus-soup/testkit/node_time.go
```diff
@@ -0,0 +1,103 @@
...
+//
+// All of this happens in a background goroutine.
+func (n *LotusNode) SynchronizeClock(ctx context.Context, genesisTime time.Time, localEpochStartCh <-chan abi.ChainEpoch, globalEpochStartCh chan<- abi.ChainEpoch) {
+	// replace the clock, setting it to the genesis timestamp + 1 epoch
+	// we are now ready to mine (and receive) the first block.
+	n.MockClock = clock.NewMock()
+	build.Clock = n.MockClock
+
+	// start at genesis + 100ms.
+	n.MockClock.Set(genesisTime.Add(100 * time.Millisecond))
```
schomatis, line 52: This still triggers (after last set of patches) the `mined block in the past` and related. In total we are advancing `BlockDelaySecs` plus this slight drift, putting us outside of the mining epoch window: the miner only advances `BlockDelaySecs` from its tipset and sees that it lost the mining window by 100ms. Ideally this should be a subtraction but in my local tests this breaks other assumptions (understandably because I'm drifting before genesis time). Maybe we should jump a smaller amount than `BlockDelaySecs` the first time (to remain inside the mining window when adding the full `BlockDelaySecs` amount in subsequent rounds).

schomatis:
I'm going to give it an in-depth review and run tomorrow but the new logic decoupling of this last iteration is beautiful :art:

lotus-soup/testkit/role_miner_sync.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/specs-actors/actors/abi"
+	"github.com/testground/sdk-go/sync"
+)
+
+// MinerSyncTopic is the topic where miner events are exchanged.
+var MinerSyncTopic = sync.NewTopic("miner_sync", &MiningSyncMsg{})
+
+// MiningSyncMsg represents a miner event.
```
schomatis, line 15:
```suggestion
// MiningSyncMsg signals that the miner attempted to mine in a particular epoch (whether successful or not).
```

Feel free to change the wording, what I'd like to achieve is more specificity in the description, this event is triggered after running `mineRound`/`MineOne`.

lotus-soup/testkit/role_miner_sync.go
```diff
@@ -0,0 +1,169 @@
...
+					continue
+				}
+
+				// we've now ascertained that all miners have reported.
+				//
+				// if this is a null round, wait for no HEAD changes.
+				// otherwise, we wait for blocks to propagate and for our HEAD
+				// to progress to a tipset that contains the blocks from all
+				// miners that produced a block, at the chain height we expect.
+				if len(produced) > 0 {
```
schomatis, line 122: Can we encapsulate the contents of this `if` in a separate function? It seems coherent enough to make sense on its own, making this `case` more compact in the containing function and possibly simplifying some of the `func()`/`defer` and `Change:`/`continue Change` artifacts.

lotus-soup/testkit/role_miner_sync.go
```diff
@@ -0,0 +1,169 @@
...
+								}
+								for _, b := range blocks {
+									if _, ok := produced[b.Miner]; !ok {
+										continue Change // not the HEAD we're looking for.
+									}
+								}
+								// the HEAD tipset includes all blocks mined in this round; we're done: release!
+								return
+							}
+						}
```
schomatis, line 148: Arriving at the end of this `for` looks like an error that should be signaled (we didn't find all the mined blocks reported).

lotus-soup/testkit/node_time.go
```diff
@@ -0,0 +1,103 @@
...
+		id = fmt.Sprintf("%s_%d", n.t.TestGroupID, n.t.GroupSeq)
+
+		minerCnt  = n.t.IntParam("miners")
+		clientCnt = n.t.IntParam("clients")
+		totalCnt  = minerCnt + clientCnt // miner is running a worker and a mining node.
+
+		epoch         = abi.ChainEpoch(0) // current epoch
+		epochInterval = time.Duration(build.BlockDelaySecs) * time.Second
+
+		ready []string
```
schomatis, line 64: Document please, it's used as a state throughout the function so seems relevant enough.

lotus-soup/testkit/node_time.go
```diff
@@ -0,0 +1,103 @@
...
+	)
+
+	go func() {
+		defer close(globalEpochStartCh)
+
+		for {
+			select {
+			case <-localEpochStartCh:
+				// move the local clock forward, and announce that fact to all other instances.
+				n.MockClock.Add(epochInterval)
```
schomatis, line 77: `MockClock` and `epoch` manipulation look like they could be encapsulated together.
For now we advance the clock _only_ to reflect a new epoch. It would make the logic easier to follow (the reader should only see the epoch change and delve into the internals to check how it's done: advancing the clock).

Not a priority, feel free to leave it as a `FIXME` if it makes sense to you.

lotus-soup/testkit/role_miner.go
```diff
@@ -353,86 +355,51 @@ func (m *LotusMiner) RunDefault() error {
...
+	if t.StringParam("mining_mode") != "natural" {
+		gen, err := m.FullApi.ChainGetGenesis(context.Background())
+		if err != nil {
+			return fmt.Errorf("failed to get genesis: %w", err)
+		}
+
+		var (
+			genTime             = time.Unix(int64(gen.MinTimestamp()), 0)
+			localEpochAdvanceCh = make(chan abi.ChainEpoch, 128)
+			globalEpochStartCh  = make(chan abi.ChainEpoch, 128)
```
schomatis, line 377: I may be reading this wrong, it's a bit confusing. There seems to be some discrepancy in the names here and in the functions arguments, like some of the `Start` should be `Advance`. (This might be intended and I'm missing something, in that case could you add some extra clarifications please?)

lotus-soup/testkit/node_time.go
schomatis: I think the simplest solution here would be to advance the clock _after_ the mining round and not before as it seems to be done now. (I think the change would translate to something like moving the clock advance logic from the local to the global case in `SynchronizeClock`.)

raulk: That would make sense. I did try different variants and in some cases I ended up hitting the "got a block from the future" error. I think we do need to get this _absolutely right_, so I'm willing to invest time in nailing the state machine right now so that we're all happy and we can build on a solid base.

lotus-soup/testkit/role_miner_sync.go
raulk: This is much better, although `successful` implies that a false value here would be a _failure`, which it isn't. We attempted to _mine_ successfully, but we weren't elected so we didn't produce a block. Will fine-tune.

lotus-soup/testkit/role_miner_sync.go
```diff
@@ -0,0 +1,169 @@
...
+	)
+
+	// obtain our address so we can broadcast it in signal messages.
+	actorAddr, err := m.MinerApi.ActorAddress(ctx)
+	if err != nil {
+		panic("failed to obtain actor address: " + err.Error())
+	}
+
+	// function to mine a single round.
+	mineRound := func(epoch abi.ChainEpoch) (produced bool) {
```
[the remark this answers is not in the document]
raulk: Yeah, I generally dislike burying goroutine spawning within functions. Let me rejigger this.

lotus-soup/testkit/node_time.go
raulk: I think advancing the clock _just before_ the mining round won't work because of races between miners. Miners who advance their clock sooner will end up propagating what appear to be "blocks from the future" to miners who still haven't advanced their clocks. That's why I opted for this approach, as we have the assurance that all miners are in the right epoch when mining commences in concert.

After removing the drift, and testing atop [this patch](https://github.com/filecoin-project/lotus/pull/2362/files/55d88440eb5cbbd31950bdf664b50b5456372ede#diff-7fc2e947b54d13e96c3477dc4e6e6c06), we no longer have the spurious logs.

The only one we're left with is the occasional:

```
WARN	sub	sub/incoming.go:81	Received block with large delay 25 from miner t01001	{"req_id": "a305c648"}
```

Which I think has to do with gossipsub propagation, and not worth investigating right now IMO.

lotus-soup/testkit/role_miner_sync.go
```diff
@@ -0,0 +1,169 @@
...
+		// mining can fail, so we perform 100 retries.
```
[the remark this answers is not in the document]
raulk: I don't really know why it can fail. I expanded a little bit, but was unable to explain under which circumstances this could happen. @yusefnapora, do you recall what the error said when this happened?

lotus-soup/testkit/role_miner_sync.go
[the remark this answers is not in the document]
raulk: This is good to go now.

lotus-soup/testkit/role_miner_sync.go
raulk: The issue is that we don't know when to throw our hands up in the air. I added a context, but it's currently mapped to context.Background(). If things don't add up, things will stall for now. This should change in the future, as we'll want to introduce a predicate so that we can simulate faults and forks.

lotus-soup/testkit/node_time.go
raulk: Done.

lotus-soup/testkit/node_time.go
raulk: I'm not sure I understood your suggestion entirely, but I think the latest iteration is way cleaner. I've modified the signature of `SynchronizeClock` to own and return the signal channels.

lotus-soup/testkit/role_miner_sync.go
```diff
@@ -0,0 +1,185 @@
...
+// signals:
+//
+//  * an epoch start signal attempts to locally mine a block, and publishes
+//    a message on a sync topic indicating that we're done mining, and whether
+//    we produced a block or not.
+//  * all miners wait until their HEAD advances to the tipset that includes
+//    all blocks that were produced in this epoch.
+//  * once that happens, we emit an epoch end signal, which the caller can
+//    interpret as a signal to advance the clock and begin a new epoch.
+func (m *LotusMiner) SynchronizeMiner(ctx context.Context, epochStart <-chan abi.ChainEpoch, epochEnd chan<- abi.ChainEpoch) {
```
schomatis, line 39: `epochStart`:
* what do we start? Mining. It seems basically a "go mine" signal.
* epoch is confusing, it may seem to imply that we mine on the epoch passed in the channel when in fact the epoch is used just for debugging purposes
* the channel seems to be doing to things at once: signal to mine (this could just be an empty struct) and _also_ tell you in which epoch the miner is standing on (do we need to obtain that information from here?)

schomatis: `epochEnd`: similar, do we even read the epoch we're sending? or is this just i'm done syncing, the terms epoch and end seem to imply to me _we are controlling the clock here_ but i think this is just a signal that i (miner) did something (either mined or synced the chain, unrelated to start/end epoch)

lotus-soup/testkit/node_time.go
```diff
@@ -0,0 +1,120 @@
...
+// The second channel is where the global clock synchronizer signals every time
+// the entirety of test participants have progressed to the next epoch, thus
+// indicating that all nodes are eady to commence that chain epoch of
+// computation.
+//
+// Essentially this behaviour can be assimilated to a distributed sempaphore.
+//
+// This method will return immediately, and will spawn a background goroutine
+// that performs the clock synchronisation work.
+func (n *LotusNode) SynchronizeClock(ctx context.Context) (localAdvance chan<- abi.ChainEpoch, globalEpoch <-chan abi.ChainEpoch, err error) {
```
schomatis, line 45: global vs local: i have trouble grappling with the topology implied by this locality

`the user requests local advancement`: it doesn't seem to me the user (miner) requests advancement, it just says 
i've synced to this epoch, yes, you can interpret it as "i'm done let's move one", but the local/global implies to me some equality that is differentiated between how we order the group, but here we have miners and a central time/chain _control_, the miners don't have any leverage on the control (in more advance use cases even if the miner says i'm done in this round we may still not advance until some other event happens), i don't see the relation that could classify them into local/global with a centralized control system

lotus-soup/testkit/role_client.go
```diff
@@ -140,6 +140,23 @@ func PrepareClient(t *TestEnvironment) (*LotusClient, error) {
...
+		}
+
+		go func() {
+			// for now, clients are always one epoch ahead locally with regards to miners.
+			// keeping the flywheel spinning!
+			for epoch := range globalEpochCh {
+				localAdvanceCh <- epoch + 1
+			}
+		}()
+	}
```
schomatis, line 158: I don't follow the rationale behind this advancement from the comment.

lotus-soup/testkit/role_miner_sync.go
raulk: Basically this routine is intended to behave like an "actor", as in the actor-oriented programming model programming. It receives signals to work on `epochStart` and emits "work done" signals on `epochEnd`. In Akka, you would assimilate the `epochStart` channel to an inbox, and `epochEnd` to an outbox (using their mailbox analogy).

I haven't thought too much about how these epochs are actually used; I started making some of these channels of element type `struct{}`, but then I realised that I might as well propagate useful info. Although in hindsight, I see how that decision (and loose usage of the sent/recvd values) might confuse others.

How about we name these channels `mineStart` and `mineDone`? From the viewpoint of this routine, it establishes the clear boundary of signals that demarcate each unit of work.

raulk:
@schomatis Good job catching the fact that it's the client that reports the "Received block with large delay" message. In general, we need to revisit the client's behaviour in all this synchronisation. I'm not convinced by the current solution.

The decision I had to take was: "do I want for the client to wait until its HEAD is updated, just like miners do?" It seemed a bit too heavy-handed to me, so that's why I decided not to wait for that condition right now, because this PR is just the first step to get fully deterministic tests (a) working [see #146], (b) doing useful stuff (we will need to simulate faults, forks, mining halting, etc.)

Right now the advancement logic is hardcoded, but to enable (b), we need some richer semantics/API where we can attach dynamically-evaluated predicates such that:

1. Miners can decide to opt-out of a mining round (advance the clock but do not attempt to mine a block).
2. Clients can watch the mining happen passively, then 1000 tipsets into the chain, decide to enroll in the ensemble, forward their clock to the current time, bootstrap against the network, and sync.
3. etc.

In conclusion, I'm not so worried about getting _everything_ right at this precise instance, because this PR is just laying the groundwork and we'll iterate on this code a lot as we introduce those predicates, checkpoints, etc.

lotus-soup/testkit/node_time.go
raulk: `SynchronizeClock` is another actor (actor-oriented programming). It encapsulates all the control of the clock. No other process in the system (not even the `SynchronizeMiner` actor) is entitled to change the clock (although we don't enforce that prohibition). The idea behind this is that when the local node, when enlisted predicates are satisfied, decides that it's ready to advance the clock, it would send a signal to `localAdvance`.

Right now we ACTUALLY advance the local clock and advertise that fact to the world. In the future, we could use some form of 3PC (three phase commit), where we PREPARE [phase 1] the advancement, nodes opt in or opt out of that round via the sync service, and once all nodes have PREPARED, we actually COMMIT [phase 2] the local clock advancement by manipulating the mock clock, and CERTIFY [phase 3] that we are at the new epoch.

Once all nodes have passed the `CERTIFY` phase, that's when we release the signal on `globalEpoch` to trigger mining and/or other downstream processes.

lotus-soup/testkit/role_miner_sync.go
schomatis: Those names sound good.

lotus-soup/testkit/node_time.go
schomatis: Yes, I think I was having the 3PC model you're describing in my mind. I'm fine if this is not implemented right now.

schomatis:
> In conclusion, I'm not so worried about getting everything right at this precise instance, because this PR is just laying the groundwork and we'll iterate on this code a lot as we introduce those predicates, checkpoints, etc.

I don't fully understand all the details here so I still don't have complete confidence in this implementation (that is on me, not an issue with the code) but I likely won't have time to review anything else today so if this is blocking the ongoing effort feel free to go ahead and merge this, we can iterate on this as we go.

lotus-soup/init.go
```diff
@@ -15,8 +15,8 @@ import (
 )
 
 func init() {
-	build.BlockDelaySecs = 2
-	build.PropagationDelaySecs = 1
+	// build.BlockDelaySecs = 2
+	// build.PropagationDelaySecs = 1
```
raulk, line 19: Need to sort this out. Synchronised mining doesn't need to reduce the block delay and propagation delay.

PR #144, opened 2020-07-14
Chain state presentation

yusefnapora:
@nonsense this builds on your branch to pull the chain state into structs that we can marshal as both plain text and json. It writes a newline-delimited `chain-state.ndjson` with a line for each epoch, as well as the state files you were already writing. I think we should keep those since they're nice and human readable and good for quick debugging.

I didn't have much time to play with the data & load into jupyter, so I don't know if the shape of the json is exactly what we want yet. But we can always tweak it.

### DIFF
--- a/lotus-soup/rfwp/chain_state.go
+++ b/lotus-soup/rfwp/chain_state.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"bytes"
 	"context"
-	"encoding"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


lotus-soup/rfwp/chain_state.go
```diff
@@ -98,110 +128,214 @@ func ChainState(t *testkit.TestEnvironment, m *testkit.LotusMiner) error {
...
+	Faults *ProvingFaultState
+	ProvingInfo *ProvingInfoState
+	Deadlines *ProvingDeadlines
+	Sectors *SectorInfo
+}
+
+// writeText marshals m to text and writes to w, swallowing any errors along the way.
+func writeText(w io.Writer, m plainTextMarshaler) {
+	b, err := m.MarshalPlainText()
+	if err == nil {
```
nonsense, line 148:
```suggestion
	if err != nil {
```

PR #145, opened 2020-07-15
message pool spam attack

vyzo:


lotus-soup/spam_attack.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+		msg := &types.Message{
+			From:     cl.Wallet.Address,
+			To:       cl.Wallet.Address,
+			Value:    one,
+			GasLimit: 10000,
+			GasPrice: one,
+		}
+		_, err = client.MpoolPushMessage(ctx, msg)
```
nonsense, line 50: Reading the code, I guess that this is testing that spamming has no effect on the time it takes to seal/retrieve content.

---

Just an idea:

I think what we also want to test is that messages with higher gas prices are never evicted from the mempool by messages with lower gas price. Hence if we have a way of monitoring the messages in the mempool (and we know we send 10 messages with gasprice 2, and a lot of messages with gasprice 1, we could periodically check that we always have the 10 messages with the gasprice 2, or that they have been mined.

It might even be worth disabling the storage miner as per https://github.com/filecoin-project/oni/pull/134 so that we know that msgs are actually in the mempool (and if they disappear from the mempool, we don't have to confirm that they've been mined)...

raulk: > I think what we also want to test is that messages with higher gas prices are never evicted from the mempool by messages with lower gas price.

There's a unit test called `TestMessageFiltering` that should be testing that at the micro level. I think our Oni tests should revolve around actually carrying out at spam attack and recording the downstream effects in terms of other system messages that are displaced as a result, e.g. proof submissions, fault submissions, recovery submissions, etc.

In that regard, I like the "spammer" role because we can mix it in with your tests, @nonsense, and we can do an A/B comparison of how the system behaves when all things are smooth sailing vs. when all hell breaks loose via a message spam attack.

@yusefnapora's plots should help us spot the change in behaviours (#147).

How does this sound?

nonsense: @raulk I agree that this is probably better suited for a unit test. I can't tell right away if the `TestMessageFiltering` is about eviction or just selecting messages for a block (messages that are already in the pool), this is why I suggested we also consider the eviction test.

raulk: @nonsense gotcha! Good point. Today I'm going to be adding a bunch of hooks in the system to [write journal entries](https://github.com/filecoin-project/lotus/pull/2101/files#diff-e92de16f120d132cacd0a0d647d5d219R328) to trace things that happen. One hook I can add trivially is mpool additions and evictions. Then we can have @yusefnapora write a nifty Jupyter widget to visualise those ;-)

PR #147, opened 2020-07-15
View chain state in jupyter

yusefnapora:
This is a first pass at analyzing chain information using juptyer / pandas. I haven't got it wired up to the jupyter / docker thing I made for the configuration UI yet, so for the moment you'll need to create a python3 virtualenv and run `pip install -r requirements.txt` in the `composer` dir. Then run `jupyter notebook` and open the `chain-state.ipynb` and either put the `chain-state.ndjson` file from a test run in the same directory or edit the `STATE_FILE` var in the notebook.

So far I've got a thing where you can pick a variable to see it plotted as a line chart:

chart-selector

And the same charts in a big stack, since maybe that's easier to glance at:

stacked-charts

I also figured out how to make a stacked area chart of the miner power distribution:

power-distribution

The state dump I'm using for the screenshots is just a few hundred blocks where the power never changes, but it should look cool with more interesting data :)


composer/chain-state.ipynb
```diff
@@ -0,0 +1,174 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
```
nonsense, line 8: Where do you edit this source? I imagine for now it happens in Jupyter and we somehow export it? Asking, because I see this is Python that is somehow JSON marshalled.

yusefnapora: I wrote this in the Jupyter UI, which isn't the nicest editor but is good for exploring. I think I'll pull it into a separate `.py` file and import it into the notebook soon though, since it's hard to read diffs of python in JSON.

PR #157, opened 2020-07-21
initial fast retrieval

nonsense:
This PR is adding the `fast_retrieval` flag as part of the manifest, so that we can run baseline deals end-to-end test with fast or non-fast retrieval.

FOLLOW UP TODOS:
- [ ] make sure we have FIL in the payment channel, so that we don't block and wait for payment when we want to measure retrieval
- [ ] how to clear CIDs from the blockstore?
- [ ] try larger sectors? 8MB sounds like a good idea - https://github.com/jimpick/lotus/commit/e6f90029ec2d841c060ba435bd32967cb2fd915e
- [ ] maybe try to inspect traces with jaeger (in case the code is instrumented) to get better understanding of what takes time during retrieval?


jimpick:
Here's an example of how to enable 8MiB sectors in Lotus:

https://github.com/jimpick/lotus/commit/e6f90029ec2d841c060ba435bd32967cb2fd915e

Here, I've modified the 2KiB debug target to use 8MiB instead.

PR #162, opened 2020-07-27
Read a subset of filecoin state over the full node API

willscott:


fc-extract-delta/lib/staterootfilter.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, act := range allActors {
+		if _, ok := goodActors[act]; ok {
+			continue
+		}
+
+		if err := tree.DeleteActor(act); err != nil {
```
raulk, line 45: We ended up going _that_ way ;-)

fc-extract-delta/lib/staterootfilter.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	addresses := make(map[address.Address]struct{})
+	populateFromTrace(&addresses, &trace.ExecutionTrace)
+	return addresses, nil
+}
+
+func populateFromTrace(m *map[address.Address]struct{}, trace *types.ExecutionTrace) {
```
raulk, line 73: maps are passed by reference in golang, so no need for the pointer here: 

* https://play.golang.org/p/AWchSOsnU6b
* https://stackoverflow.com/questions/40680981/are-maps-passed-by-value-or-by-reference-in-go

fc-extract-delta/main.go
```diff
@@ -0,0 +1,178 @@
...
+	headers := http.Header{}
+	if len(token) != 0 {
+		headers.Add("Authorization", "Bearer "+string(token))
+	}
+	node, _, err := client.NewFullNodeRPC(addr, headers)
+	if err != nil {
+		return nil, fmt.Errorf("could not connect to api: %w", err)
+	}
+	return node, nil
+}
```
nonsense, line 89: Heads up: at https://github.com/filecoin-project/oni/pull/167/files#diff-3fc1d1cd75e2088d9da8d680c52a0458R30-R63 I am using directly a `repo` and `BlockStore` and `ChainStore` - I noticed that most of the API methods related to the blockchain are pretty much forwarded to the `ChainStore`, so I haven't had a need for `api.FullNode` so far.

This makes it easier to just point to the Lotus repo directory, even if the node is not running.

willscott: seems easy enough to abstract out. the benefit of `api.FullNode` is that interface is the thing that should be consistent for other implementatiuons, while the on-disk format of `BlockStore`/`ChainStore` may be lotus specific.

raulk:
Work incorporated in https://github.com/filecoin-project/oni/pull/177 (merged).

Issue #166, opened 2020-07-28
implementation independent progressive three or N-way state diffs (MVP)


raulk:
# Context

Our test vectors are expected to carry:

- `initialisation`: pre-seed state, as a map of address to actor state that we need to seed into a fresh, empty state tree to initialise the environment for the test.
- `applies`: items that the VM is asked to apply (messages, blocks, tipsets).
- `expectations`: what we expect the world to look like after the applies are done.
  - a state tree root CID.
  - full state tree HAMT for debugging purposes (this can get heavy quickly, for MVP purposes it's fine to embed the full output state tree in the test vector; in the future, we may want to capture only the diff with respect to the initialisation tree).

There are a minimum of three state trees involved in a given test:
- the input state tree.
- the actual output state tree (as produced by the VM under test).
- the expected output state tree (as embedded in the state vector).

Some test vectors may contain multi-stage expectations. For example, chain-class test vectors could carry output state trees after every tipset. That means that there may be more than 3 state trees to juggle during the execution of a test vector.

# Goal

Build a MVP CLI tool that takes two serialised state trees (either as command arguments or in car? files), and produces an actor-level diff. This tool uses specs-actors to perform the interpretation of the state tree and deserialise the state into struct level objects.

We might have to end up comparing IPLD dags.

The "progressive" characteristics comes from the fact that we start coarse/shallow, to gradually get finer/deeper in our comparisons.

* First, we compare the state tree roots and report any mismatches.
* Second, we traverse the trees/HAMT and report actor state root mismatches.
* Third, we perform byte-level comparison of the serialised CBOR forms of the actor states.
* Fourth, we deserialise the CBORs into the state structs (using specs-actors), and we perform a struct-wise comparison using something like https://pkg.go.dev/github.com/google/go-cmp/cmp?tab=doc#Diff.

PR #169, opened 2020-07-28
test vector schema draft + example 'message' class test vector.

### LINKED ISSUE #163: define a VM Conformance Test Vector schema
# Goal

To define a test vector schema in JSON that can express all the necessary features of a test vector.

We foresee FOUR classes of test vectors, with a recursive encapsulation pattern.

* Message class: test a single message being applied on top of an existing state.
* Block class: test an entire block being applied on top of an existing state.
* TipSet class: test a tipset being applied on top of an existing state.
* Chain class: test a sequence of tipsets being applied on top of an existing state.

Test vectors should be expressed in a serialisation format that is intuitive and easy to understand by humans, and efficient to parse and process by machines, interoperable, ideally commentable, and that is amply supported in other programming languages like Rust and C++.

# Notes

* For simplicity, let's pick JSON, although barebones JSON fails the "commentable" requirement.
  * If push comes to shove, we could embed comments in `_` prefixed fields.
  * I like JSON because it has JSON schema, it's pervasive, and has superb cross-language support. Some things suck, but you can't have all the things.
  * Other formats considered:
    * TOML: no schema, awkward/verbose array notation.
    * yaml: too complex.
    * https://json5.org/: like JSON but with comments and a few other features. No good cross-language support.
    * HCL (Hashicorp Configuration Language): superset of JSON, pretty nice like everything Hashicorp ❤️, but poor cross-language support (mostly Go).
    * HOCON: rather obscure and JVM-oriented.
* Test vectors will be consumed by Lotus, Forest (rust-filecoin), and Fuhon (cpp-filecoin).
* Test vectors will be generated either from an existing chain, or via a DSL, never crafted manually.
* Binary data will be hex encoded.
  * Extra points if we can annotate the values with comments, so they can be understood and debugged.
  * Inspiration:
    * Think of "source maps" or "debugging information" in the world of compilers.
    * Think of how Wireshark overlays the mapping of binary values to fields in a high-level protocol data structure.
* Asserts are expected to be made at the state tree root CID level.
  * This is fine for a first iteration. However, this leads to poor debuggability.
  * We will need to produce progressive, deeper diffs upon a state tree root mismatch, so the developer can pinpoint where the problem is.
  * Compare the state roots, then compare per actor state roots, then compare the CBORs, then decode the CBORs into their structs.
  * Use a library like [go-cmp](https://pkg.go.dev/github.com/google/go-cmp/cmp?tab=doc#Diff) to render nice object-level diffs of actor state mismatch.

raulk:
Fixes #163.

Uses a declarative style.

### DIFF
--- a/test-vectors/message.json
+++ b/test-vectors/message.json
@@ -1,0 +1,32 @@
+{
+  "class": "message",
+  "selector": "<some predicate for the driver to determine if this vector applies to the implementation, e.g. to deal with protocol upgrades>",
+
+  "_meta": {
+    "id": "test_vector_id",
+    "version": "version_id",
+    "gen": {
+      "source": "lotus",
+      "version": "0.4.1+git.27d74337+api0.8.1"
+    }
+  },
+
+  "preconditions": {
+    "state_tree": {
+      "t01": "0x0123456789abcdef",
+      "t02": "0x0123456789abcdef",
+      "t03": "0x0123456789abcdef"
+    }
+  },
+
+  "apply_messages": [
+    "0x89004300ba20583103b85358cc1c968d826c0a0efcde99b6e4db138edfc051f188a4471cfc23ccf7884adbbac912e45f7363774bf16f83ca8f1902b44800505e1442e66b744200011a09d225c80758c68219016f58c099a447a8c80988e1c061dd3a8dd674033c3a6083c4b55cf035e6e8de3baf09436ad00adf6eb14d58ae6691d22ce82255874047e7aa0b9a86bbf424afc6d89dc29f7b253dd04c88a625b86345f05aa0bdee503971c1e9735ed5b81c31ad41a1f10b387d96151d1cbd0c4286267daf04087fd8d5272f870994c3732f3fc4f54287a485d6c8dafb6056df3fc9233f849027b4c1272850435126fef65a8363768a858c0cc5f9d9114ff378ad50cab25ab0a2f685ca63d9f1f7727da4d06db5b80b5f"
+  ],
+
+  "postconditions": {
+    "state_tree": {
+      "root_cid": "bafy2bzacecfk5enlkya62353pwt5wjn3v7mldibtzgohfpbjw4xmabm34evz4",
+      "car_bytes": "0x890043009920583103b5b1fab6769fcb464146d192af785915d8e1dacff264b2be3bcb7bf4064a77cb3b13e27063ea4825dc48cd0a7df77e211916e748002740f0c493a2824200011a09d9c4dc0758c68219091c58c083e7c59e748e070b382a813f4ffe8ead2a090469aaa37f0664b82c092bc78b43fac300335cd2d6e4a8c71ce138003c0fa8ee1707b41aa87f35c811613d59ac53449aef8263aa51c876f2e67c185118ec9628f73442f58e093909ce952e431e770ac693cdbe99f2f9404ad3ca754196ab443c6c66afa41310e705e453496bdcedf0f271529dc7374021d7b80d343397f996d06bf11e753a49eaed7b3ce80a8e9bbaa48bf83223527bd8a2933f002949fdb8058103336a4a979375d5e963cdd7b1"
+    }
+  }
+}
--- a/test-vectors/schema.json
+++ b/test-vectors/schema.json
@@ -1,0 +1,156 @@
+{
+  "$id": "https://filecoin.io/oni/schemas/test-vector.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "a filecoin VM test vector",
+  "type": "object",
+  "definitions": {
+    "hex": {
+      "title": "hex value",
+      "description": "a hex value prefixed with 0x, and accepting only lowercase characters; 0x represents an empty byte array",
+      "type": "string",
+      "pattern": "0x[0-9a-f]*",
+      "examples": [
+        "0xa1b2c3",
+        "0x"
+      ]
+    },
+    "meta": {
+      "title": "metadata",
+      "description": "metadata about this test vector, such as its id, version, data about its generation, etc.",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "id": {
+          "type": "string"
+        },
+        "version": {
+          "type": "string"
+        },
+        "gen": {
+          "title": "generation metadata",
+          "description": "metadata about how this test vector was generated",
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "source": {
+              "type": "string",
+              "examples": [
+                "lotus",
+                "dsl"
+              ]
+            },
+            "version": {
+              "type": "string",
+              "examples": [
+                "0.4.1+git.27d74337+api0.8.1"
+              ]
+            }
+          }
+        }
+      }
+    },
+    "preconditions": {
+      "title": "execution preconditions",
+      "description": "preconditions that need to be applied and satisfied before this test vector can be executed",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "state_tree": {
+          "title": "state tree to seed",
+          "description": "state tree to seed before applying this test vector; mapping of actor addresses => serialized state",
+          "type": "object",
+          "additionalProperties": {
+            "$ref": "#/definitions/hex"
+          },
+          "examples": [
+            {
+              "t01": "0x0123456789abcdef",
+              "t02": "0x0123456789abcdef",
+              "t03": "0x0123456789abcdef"
+            }
+          ]
+        }
+      }
+    },
+    "postconditions": {
+      "title": "execution preconditions",
+      "description": "postconditions that need to be satisfied after execution for this test vector to pass",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "state_tree": {
+          "title": "state tree postconditions",
+          "description": "state tree postconditions that must be true for this test vector to pass",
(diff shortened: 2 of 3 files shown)


test-vectors/message.json
```diff
@@ -0,0 +1,32 @@
...
+      "version": "0.4.1+git.27d74337+api0.8.1"
+    }
+  },
+
+  "preconditions": {
+    "state_tree": {
+      "t01": "0x0123456789abcdef",
+      "t02": "0x0123456789abcdef",
+      "t03": "0x0123456789abcdef"
+    }
```
raulk, line 19: @willscott this is where we'll set the plucked/masked state tree.

@nonsense this is the state tree we need to slip in the genesis block.

test-vectors/message.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+
+  "apply_messages": [
+    "0x89004300ba20583103b85358cc1c968d826c0a0efcde99b6e4db138edfc051f188a4471cfc23ccf7884adbbac912e45f7363774bf16f83ca8f1902b44800505e1442e66b744200011a09d225c80758c68219016f58c099a447a8c80988e1c061dd3a8dd674033c3a6083c4b55cf035e6e8de3baf09436ad00adf6eb14d58ae6691d22ce82255874047e7aa0b9a86bbf424afc6d89dc29f7b253dd04c88a625b86345f05aa0bdee503971c1e9735ed5b81c31ad41a1f10b387d96151d1cbd0c4286267daf04087fd8d5272f870994c3732f3fc4f54287a485d6c8dafb6056df3fc9233f849027b4c1272850435126fef65a8363768a858c0cc5f9d9114ff378ad50cab25ab0a2f685ca63d9f1f7727da4d06db5b80b5f"
+  ],
+
+  "postconditions": {
+    "state_tree": {
+      "root_cid": "bafy2bzacecfk5enlkya62353pwt5wjn3v7mldibtzgohfpbjw4xmabm34evz4",
```
raulk, line 28: @willscott in the progressive diff issue (#166), this is the first level to compare.

test-vectors/message.json
```diff
@@ -0,0 +1,32 @@
...
+      "car_bytes": "0x890043009920583103b5b1fab6769fcb464146d192af785915d8e1dacff264b2be3bcb7bf4064a77cb3b13e27063ea4825dc48cd0a7df77e211916e748002740f0c493a2824200011a09d9c4dc0758c68219091c58c083e7c59e748e070b382a813f4ffe8ead2a090469aaa37f0664b82c092bc78b43fac300335cd2d6e4a8c71ce138003c0fa8ee1707b41aa87f35c811613d59ac53449aef8263aa51c876f2e67c185118ec9628f73442f58e093909ce952e431e770ac693cdbe99f2f9404ad3ca754196ab443c6c66afa41310e705e453496bdcedf0f271529dc7374021d7b80d343397f996d06bf11e753a49eaed7b3ce80a8e9bbaa48bf83223527bd8a2933f002949fdb8058103336a4a979375d5e963cdd7b1"
```
raulk, line 29: @willscott in the context of the progressive diff issue (#166), this contains a dump of the expected resulting state tree. In the future, we might want to switch this to a diff against the initial state tree, for conciseness and space efficiency.

test-vectors/message.json
```diff
@@ -0,0 +1,32 @@
...
+
+  "preconditions": {
+    "state_tree": {
+      "t01": "0x0123456789abcdef",
+      "t02": "0x0123456789abcdef",
+      "t03": "0x0123456789abcdef"
+    }
+  },
+
+  "apply_messages": [
```
raulk, line 22: Depending on the test vector class, the JSON schema will enforce one of the following fields:

* `apply_message`
* `apply_block`
* `apply_tipset`
* `apply_chain`

test-vectors/message.json
raulk: This could be a serialized CAR, but I thought we might as well set it in map form, because:
1. it's more obvious, human-readable, and easier to debug.
2. we control this input entirely (as it's generated by the generators).
3. VM implementation may have custom mechanics for state imports, and this is the most explicit choice, which makes it a sane fallback/common denominator.

test-vectors/schema.json
```diff
@@ -0,0 +1,156 @@
...
+              "t01": "0x0123456789abcdef",
+              "t02": "0x0123456789abcdef",
+              "t03": "0x0123456789abcdef"
+            }
+          ]
+        }
+      }
+    },
+    "postconditions": {
+      "title": "execution preconditions",
```
alanshaw, line 76:
```suggestion
      "title": "execution postconditions",
```

test-vectors/schema.json
```diff
@@ -0,0 +1,156 @@
...
+          "description": "state tree to seed before applying this test vector; mapping of actor addresses => serialized state",
+          "type": "object",
+          "additionalProperties": {
+            "$ref": "#/definitions/hex"
+          },
+          "examples": [
+            {
+              "t01": "0x0123456789abcdef",
+              "t02": "0x0123456789abcdef",
+              "t03": "0x0123456789abcdef"
```
alanshaw, line 69: Do you see these serializations eventually being expanded out into a schema we can use to author tests solely in JSON?

test-vectors/schema.json
```diff
@@ -0,0 +1,156 @@
...
+      "title": "test vector class",
+      "description": "test vector class; depending on the value, the apply_* property to provide (and its schema) will vary; the relevant apply property is apply_[class]",
+      "type": "string",
+      "enum": [
+        "messages",
+        "block",
+        "tipset",
+        "chain"
+      ]
+    },
```
alanshaw, line 121: I understand messages can transition the state. I'm not clear on what a test vector for the other classes listed here might look like - are they just examples at this point?

How would we express a test vector that applies messages, blocks and then some more messages and then asserts on the state? ...is this too high level?

test-vectors/message.json
nonsense: I am not sure what is the best format to specify the input state tree (just because I don't have a good understanding of what exactly is the state right now) - serialised CAR would definitely work, I am not sure about individual key/values and whether it makes sense to do that. Also here you specify t01 t02 t03 - these are not CIDs, I assume you mean that this is the state of the miners, or?

test-vectors/schema.json
```diff
@@ -0,0 +1,156 @@
...
+  "properties": {
+    "class": {
+      "title": "test vector class",
+      "description": "test vector class; depending on the value, the apply_* property to provide (and its schema) will vary; the relevant apply property is apply_[class]",
+      "type": "string",
+      "enum": [
+        "messages",
+        "block",
+        "tipset",
+        "chain"
```
frrist, line 119: What are the differences between the `tipset` and `chain` cases?

test-vectors/single_message.json
```diff
@@ -0,0 +1,32 @@
...
+  ],
+
+  "postconditions": {
+    "state_tree": {
+      "root_cid": "bafy2bzacecfk5enlkya62353pwt5wjn3v7mldibtzgohfpbjw4xmabm34evz4",
+      "car_bytes": "0x890043009920583103b5b1fab6769fcb464146d192af785915d8e1dacff264b2be3bcb7bf4064a77cb3b13e27063ea4825dc48cd0a7df77e211916e748002740f0c493a2824200011a09d9c4dc0758c68219091c58c083e7c59e748e070b382a813f4ffe8ead2a090469aaa37f0664b82c092bc78b43fac300335cd2d6e4a8c71ce138003c0fa8ee1707b41aa87f35c811613d59ac53449aef8263aa51c876f2e67c185118ec9628f73442f58e093909ce952e431e770ac693cdbe99f2f9404ad3ca754196ab443c6c66afa41310e705e453496bdcedf0f271529dc7374021d7b80d343397f996d06bf11e753a49eaed7b3ce80a8e9bbaa48bf83223527bd8a2933f002949fdb8058103336a4a979375d5e963cdd7b1"
+    }
+  }
+}
```
frrist, line 32: In the event that this tests fails (meaning the resulting state root != `root_cid`(?)) what information will be given to the caller of the test? State roots can differ for a variety of reasons: different gas usage, exit codes, return values, etc. I think it will be helpful to surface why state roots are different -- is tooling being built to facilitate that?

test-vectors/schema.json
```diff
@@ -0,0 +1,156 @@
+{
+  "$id": "https://filecoin.io/oni/schemas/test-vector.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "a filecoin VM test vector",
```
frrist, line 4: nit: this will likely cover more than just the VM since these tests validate the application of tipsets and/or complete chains.

test-vectors/single_message.json
willscott: https://github.com/filecoin-project/oni/issues/166

test-vectors/schema.json
nonsense: From what I understand we are not supposed to do that - see https://docs.google.com/document/d/1_wlRp-DBVqlTijP-Jq3SzSVd65ABgnXD1pw_yBxip3I/edit?disco=AAAAKA9s15M

willscott: we should use the same definition / schema format for state_tree in preconditions and post conditions. It's extra work to store it as an array of actor states as the pre-condition, and then as a car + root in the post condition.

raulk:
Initial incorporated in https://github.com/filecoin-project/oni/pull/177 (merged).
Further schema refinement to follow in forthcoming PRs.

Issue #175, opened 2020-08-03
port chain validation builders to spit out test vectors


raulk:
## Goal

Port [the `chain` package under chain-validation](https://github.com/filecoin-project/chain-validation/tree/master/chain) to produce Oni test vectors.

In other words, use that API as a frontend, and switch up the backend so that we're writing out Oni test vectors.

nonsense:
Done in #181

PR #181, opened 2020-08-04
generate test-vectors based on tests from chain-validation project

nonsense:
This PR is addressing https://github.com/filecoin-project/oni/issues/175

It is refactoring the `chain-validation` module for the purpose of producing test vectors for Project Oni, while still keeping some of the validation and assertions when producing the test vectors.

I've left the `assert` and the `require` (`testify` packages) as sanity checks while we are generating the test vectors.

---

```
tvx suite-messages > all-message-tests
```

---

I realize this is a huge PR - it is a mash-up of the `filecoin-project/chain-validation` project and the `lotus/chain/validation` packages. I've removed some of the indirection that we won't need (such as `Factories` and a few of the wrappers (for example `Validation`)), and also tried to get rid of the Go `testing.T` usage as much as possible for a first pass.

In terms of difference of usage, you can compare:

`TestChainValidationMessageSuite` from Lotus...     vs
`MessageTest_AccountActorCreation` from this PR

TODO:
- [x] gofmt
- [x] use Lotus type `Message` and not custom type, as `exec-lotus` needs to deserialise Lotus type `Message`.
- [x] fix gzip of car

TODO AS A FOLLOW UP:
- [x] generate test vectors for the messages/create-actor suite
- [x] generate test vectors for the messages/message-application suite
- [ ] generate test vectors for the messages/multisig suite
- [ ] generate test vectors for the messages/nested suite
- [ ] generate test vectors for the messages/paych suite
- [ ] generate test vectors for the messages/transfer suite
- [ ] generate test vectors for the tipset suite


nonsense:
I see this is already conflicting with `tvx` branch - will address conflicts tomorrow!

In any case a quick / skim review on the approach/direction would be appreciated ;)

tvx/chain/types/message.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/filecoin-project/specs-actors/actors/abi"
+	"github.com/filecoin-project/specs-actors/actors/abi/big"
+	crypto "github.com/filecoin-project/specs-actors/actors/crypto"
+	block "github.com/ipfs/go-block-format"
+	"github.com/ipfs/go-cid"
+	"github.com/multiformats/go-multihash"
+	cbg "github.com/whyrusleeping/cbor-gen"
+)
+
+type Message struct {
```
willscott, line 18: do we need to import this? can we reference the `lotus/chain/types` variant directly?

tvx/drivers/applier.go
```diff
@@ -0,0 +1,151 @@
...
+	lotusVM, err := vm.NewVM(base, epoch, &vmRand{}, a.stateWrapper.bs, a.syscalls, nil)
+	// need to modify the VM invoker to add the puppet actor
+	chainValInvoker := vm.NewInvoker()
+	chainValInvoker.Register(puppet.PuppetActorCodeID, puppet.Actor{}, puppet.State{})
+	lotusVM.SetInvoker(chainValInvoker)
+	if err != nil {
+		return vtypes.MessageReceipt{}, big.Zero(), big.Zero(), err
+	}
+
+	ret, err := lotusVM.ApplyMessage(ctx, lm)
```
willscott, line 129: we may want to eventually keep the VM open across multiple actions rather than flushing between each one.

more generally, we probably want to factor out the use of lotusVM to be able not directly run our linked version of the VM in-process?

tvx/drivers/dummy_testingt.go
```diff
@@ -0,0 +1,28 @@
+package drivers
+
+import (
+	"fmt"
+	"os"
+)
+
+var t dummyTestingT
+
+type dummyTestingT struct{}
```
willscott, line 10: can we remove this? it seems like a cludge - panic directly where this would panic, rather than making the code look like it's been imported from a test

nonsense: There are probably >100 places that use an interface subset of testing.T (all the `require` and `assert` statements). I'd rather not change all that right now, as that would make the PR twice as big.

I know this is ugly, but I didn't think of a better way, and updating all the code to get rid of `testify` framework doesn't seem like a good use of our time right now.

tvx/chain/types/message.go
nonsense: I am not sure - there are different fields (for example `CallSeqNum`, and `Version`), and there is a test that we should migrate which is testing `CallSeqNum`.

tvx/drivers/applier.go
nonsense: I see your point about flushing on every action (i.e. on every message) - this is probably redundant... although I'd consider this an optimisation - the result should be the same, no matter if we keep one VM, or we create a new one based on the same state. Maybe the VM could be part of the `stateWrapper`, but for now I didn't want to introduce more than the bare minimum of changes.

I am not sure I follow what you mean with `to run our linked version of the VM` though?

tvx/drivers/dummy_testingt.go
willscott: just flagging that this is a lot of technical debt we're taking on at the start of this and now are gonna have to be maintaining :-|

tvx/drivers/applier.go
willscott: this code dep's to a specific version of lotus & spec actors, and we're going to be tied to that as our reference as that's the code we'll be directly executing when we call vm here. Probably an pre-optimization worry that isn't a big deal yet.

nonsense:
> the thing i'm most worried about here is that there's a lot of code here that appears implementation specific - sufficiently so that i'm scared small bits of this will continually diverge from changes in lotus, and we're stuck updating this to stay in sync. Would be great to catalog the different parts enough to understand what we can try to re-use from upstream rather than doing ourselves.

I see your point, this happened to some extend with the `lotus-soup` test plan, and we want to have a better experience this time around. On the other hand, even though there is a lot of code in this PR, I don't think we will have that problem, because the part that we are using are mostly from `specs-actors` and they should be stable by now. There is really not that much that is implementation specific, but we should definitely document it.

- `Message`, `SignedMessage` types/serialisation
- message `Receipts` and `Gas` types
- sector types (`PreSeal`)
- wallet/key types (`secp256k1`, `bls`)

- sugar code around builtin `spec-actors` methods - I imagine new ones might come up, and we need to be aware to test them (multisig, paychan, power, reward, etc.)

Lotus:
- `IpldStore`, `StateManager`, `blockstore`, `datastore` - these might change, but no way around not using them.
- `vm`
- `syscall` functions (chain validation) - I imagine these are stable by now.
- a few consts (total Filecoin, precision)

---

Having enumerated what I think are some high-level dependencies to Lotus and specs-actors, probably a better measure of how often these things change is to monitor the `chain-validation` project - it has to be amended when specs-actors/Lotus changes, and this does seem to happen occasionally.

tvx/drivers/applier.go
nonsense: I think it is fine to use the Lotus VM for now. I don't think we have in scope to implement a reference VM at this stage, so no way around not including specs-actors and Lotus VM and living with them changing under our feet (which at this stage really shouldn't happen often, or should be minimal).

I suggest we add an abstraction on the VM and inject it as soon as we have another impl. but not before.

tvx/chain/types/message.go
nonsense: Actually you are right about that. `exec-lotus` is trying to deserialise the `types.Messages`, so this is what the test vector should marshal. Will update and see if we can still keep some of the tests for the nonce applicable.

nonsense: Added as a TODO. Will address.

willscott:
are you going to get all of the things in your top post done before merging this, or is there a checkpoint before that where we can merge this PR and help mitigate against continuing to resolve merge conflicts / divergence?

nonsense:
> are you going to get all of the things in your top post done before merging this, or is there a checkpoint before that where we can merge this PR and help mitigate against continuing to resolve merge conflicts / divergence?

I think the `TODO` list is a must, the rest should be done as follow up, as this PR is getting into conflicts every 5min.

willscott:
> > are you going to get all of the things in your top post done before merging this, or is there a checkpoint before that where we can merge this PR and help mitigate against continuing to resolve merge conflicts / divergence?
> 
> I think the `TODO` list is a must before merging, the rest should be done as follow up, as this PR is getting into conflicts every 5min. WDYT?

Getting the same gzip / format as in master right now and then merging and doing the others as separate PRs seems preferable :+1

raulk:
> > > are you going to get all of the things in your top post done before merging this, or is there a checkpoint before that where we can merge this PR and help mitigate against continuing to resolve merge conflicts / divergence?
> > 
> > 
> > I think the `TODO` list is a must before merging, the rest should be done as follow up, as this PR is getting into conflicts every 5min. WDYT?
> 
> Getting the same gzip / format as in master right now and then merging and doing the others as separate PRs seems preferable :+1

Agree. Let's merge ASAP so we can divide and conquer amongst the team.

nonsense:
> Just wanted to point that this type seems to add nothing on top of https://github.com/filecoin-project/lotus/blob/7f722623caa95281b22e66f5f09c3d0faf198d16/chain/types/message.go#L27, unless it's here for the purposes of adding custom methods to it, in which case we can consider alternative ways that do not require copying the entire struct.

Removed it.

tvx/dummy_testingt.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import (
+	"fmt"
+	"os"
+)
+
+var t dummyTestingT
```
willscott, line 8: this is duplicated both in top level and in chain/ ?

nonsense: Removed the top-level one - we now only have `require` and `assert` statements in `chain`.

PR #185, opened 2020-08-05
support multiple messages /1

nonsense:
This works only within a single epoch... I imagine if we want to apply blocks (and change epoch), we wouldn't use this test vector class.


tvx/exec_lotus.go
```diff
@@ -77,16 +78,25 @@ func runExecLotus(_ *cli.Context) error {
...
-		}
-
 		driver := lotus.NewDriver(ctx)
 
-		fmt.Println("executing message")
-		spew.Dump(driver.ExecuteMessage(msg, header.Roots[0], bs, epoch))
+		root := header.Roots[0]
+
+		for i, m := range tv.ApplyMessages {
+			fmt.Printf("decoding message %v\n", i)
```
willscott, line 86: prefer log or printing to stderr. here and below

raulk:
Schema needs to be updated, but this LGTM.

tvx/exec_lotus.go
nonsense: Let's address as a follow up, as we have many println and spew all over the codebase.

PR #186, opened 2020-08-05
parse multiple test cases per file /2

nonsense:
In https://github.com/filecoin-project/oni/pull/181 , I am generating multiple test vectors per file, that are new line delimited.

This is adding support to actually process all of them, and not just one per file.

### DIFF
--- a/tvx/exec_lotus.go
+++ b/tvx/exec_lotus.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 
 	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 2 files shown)


raulk:
Actually, let me retract my comment above. On thinking about it a bit more, and in order to support https://github.com/filecoin-project/oni/issues/189, I think generating multiple test vectors from a single go test vector builder script makes a lot of sense!

raulk:
So yet another refinement on my thinking. I think `exec-lotus` should accept:

* a file containing a single test vector: `--file`
* a directory, containing one or multiple test vectors: `--dir`
* stdin containing a stream of ndjson (newline delimited json) vectors: `--stdin`

But _not_ a single file containing multiple test vectors.

The stdin will work really well if we make the builder scripts output ndjson on stdout, like I suggested in https://github.com/filecoin-project/oni/issues/189. Then we can do:

```sh
$ go run gen/my_suite/mario.go | tvx exec-lotus --stdin
```

To generate a bunch of vectors and execute them in Lotus.

PR #187, opened 2020-08-05
dedup state tree CARs from test vectors.

### LINKED ISSUE #184: deduplicate state trees in test vectors
Currently we are storing the pre and post trees in independent fields as one separate CAR each. Since executing a test vector is supposed to alter a relatively small part of the state tree (affected leaves + all their paths to the root), deduplication will be highly efficient.

After considering several solutions within the team, we have arrived at the following implementation. Instead of encoding the pre and post trees as independent CARs, embed a single base64-encoded CAR in the test vector in a top-level field, compressed with Brotli/gzip (see #183). The preconditions and postconditions blocks would reference different root CIDs from within the single CAR.

In the future, we may want to expand the test vector schema to host multiple CARs, but we don't see the use case for this as yet. Will cross that bridge if/when we get to it.

## Tasks

- [ ] Modify the test vector schema and structs.
- [ ] Adapt existing code.

raulk:
Fixes #184.


corpus/single_message.json
```diff
@@ -11,19 +11,24 @@
...
+      "root_cid": {
+        "/": "bafy2bzacebbxsepazfgwepawspvzenb2x64pmqjyan3wgtfpxu5nxez33wzkc"
+      }
     }
   },
 
   "apply_message": "0x89004300ba20583103b85358cc1c968d826c0a0efcde99b6e4db138edfc051f188a4471cfc23ccf7884adbbac912e45f7363774bf16f83ca8f1902b44800505e1442e66b744200011a09d225c80758c68219016f58c099a447a8c80988e1c061dd3a8dd674033c3a6083c4b55cf035e6e8de3baf09436ad00adf6eb14d58ae6691d22ce82255874047e7aa0b9a86bbf424afc6d89dc29f7b253dd04c88a625b86345f05aa0bdee503971c1e9735ed5b81c31ad41a1f10b387d96151d1cbd0c4286267daf04087fd8d5272f870994c3732f3fc4f54287a485d6c8dafb6056df3fc9233f849027b4c1272850435126fef65a8363768a858c0cc5f9d9114ff378ad50cab25ab0a2f685ca63d9f1f7727da4d06db5b80b5f",
 
   "postconditions": {
     "state_tree": {
```
willscott, line 28: can we have pre/post just directly encode root_cid now? the `state_tree` layer seems unneeded now.

raulk: We could flatten this to `state_tree_root_cid`, but I'd prefer to keep the nesting just in case we need to add more elements to the `state_tree`. We still have unknown unknowns here. Once all the pieces have fit together entirely, we can come back and dry out the schema.

PR #207, opened 2020-08-06
fix two disabled tests

nonsense:
This PR is enabling the two tests I disabled yesterday since they were failing.

When executing a message, the fact that we can't find an actor doesn't mean we should not execute the message... or does it?

### DIFF
--- a/tvx/lotus/driver.go
+++ b/tvx/lotus/driver.go
@@ -32,10 +32,10 @@ func (d *Driver) ExecuteMessage(msg *types.Message, preroot cid.Cid, bs blocksto
 
 	actor, err := st.GetActor(msg.From)
 	if err != nil {
-		return nil, cid.Undef, err
+		fmt.Println("from actor not found: ", msg.From)
+	} else {
+		fmt.Println("from actor found: ", actor)
 	}
-
-	fmt.Println("from actor found: ", actor)
 
 	fmt.Println("creating vm")
 	lvm, err := vm.NewVM(preroot, epoch, &vmRand{}, bs, mkFakedSigSyscalls(vm.Syscalls(ffiwrapper.ProofVerifier)), nil)
--- a/tvx/messages_message_application.go
+++ b/tvx/messages_message_application.go
@@ -52,26 +52,25 @@ func MessageTest_MessageApplicationEdgecases() error {
 	}
 
 	err = func(testname string) error {
-		//TODO: this test is broken, fix later
-		return nil
 		td := drivers.NewTestDriver()
 
 		v := newEmptyMessageVector()
-
 		alice, _ := td.NewAccountActor(drivers.SECP, aliceBal)
 
 		preroot := td.GetStateRoot()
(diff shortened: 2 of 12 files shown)


tvx/messages_message_application.go
```diff
@@ -52,26 +52,25 @@ func MessageTest_MessageApplicationEdgecases() error {
...
+		v.ApplyMessages = append(v.ApplyMessages, chain.MustSerialize(msg))
+
 		// Expect Message application to fail due to lack of gas
 		td.ApplyFailure(
 			msg,
 			exitcode_spec.SysErrOutOfGas)
 
-		// Expect Message application to fail due to lack of gas when sender is unknown
 		unknown := chain.MustNewIDAddr(10000000)
 		msg = td.MessageProducer.Transfer(unknown, alice, chain.Value(transferAmnt), chain.Nonce(0), chain.GasPrice(10), chain.GasLimit(1))
```
nonsense, line 70: I am not 100% sure I understand this test. How can the sender be unknown? Isn't this messages supposed to be stopped on the mempool level, and never reach the VM?

nonsense: cc @frrist (marking you as I see you might be the original author of this test and shed some light ;)

tvx/lotus/driver.go
```diff
@@ -32,11 +32,11 @@ func (d *Driver) ExecuteMessage(msg *types.Message, preroot cid.Cid, bs blocksto
 
 	actor, err := st.GetActor(msg.From)
 	if err != nil {
-		return nil, cid.Undef, err
+		fmt.Println("from actor not found: ", msg.From)
```
willscott, line 35: this seems like a bad thing? is there a reason we want to continue when no source actor is in the VM?

nonsense: The invalid test needs to get to the testing stage basically.

tvx/messages_message_application.go
nonsense: Based on chats we've had today, it appears that such a message really should not be part of a `valid` block, but it is always good to confirm that the VMs of various clients would end up at the same state, if we throw such an invalid message to the VM.

frrist: In the event that an invalid message (like the one here) isn't filtered by block validation we need to ensure behavior is consistent across VM implementations. The assertion being made here is checking the order of validation checks in a filecoin VM:
[Lotus Validation order](https://github.com/filecoin-project/lotus/blob/3058f280d9ba34c044055e2a85a63bd07d52b307/chain/vm/vm.go#L329-L345)
[go-filecoin Validation Order](https://github.com/filecoin-project/go-filecoin/blob/8f6abbaaec2df92eb9a9ededbfc1d9dee035c987/internal/pkg/vm/internal/vmcontext/vmcontext.go#L374-L394)

raulk: Yeah, these kinds of messages would really benefit from capturing the `SysErrOutOfGas` in the new postconditions receipts.

Issue #194, opened 2020-08-05
generate message-class test vectors based on `chain-validation` project


nonsense:
As a follow up to https://github.com/filecoin-project/oni/pull/181, we should review and convert all tests into test vectors, if they make sense:

https://github.com/filecoin-project/chain-validation/tree/master/suites

TODO:
- [x] generate test vectors for the messages/create-actor suite
- [x] generate test vectors for the messages/message-application suite
- [x] generate test vectors for the messages/multisig suite
- [x] generate test vectors for the messages/nested suite
- [x] generate test vectors for the messages/paych suite
- [x] generate test vectors for the messages/transfer suite

raulk:
@alanshaw @willscott help requested here to divide and conquer! Comment if you'd like to pick up the migration of an item here, and ping @nonsense to align.

alanshaw:
I'm going to take a stab at `messages/multisig`

nonsense:
I am picking up payment channel tests (updated the description).

nonsense:
Will be picking up the messages/transfer suite tomorrow.

nonsense:
Starting `messages/nested` now.

raulk:
I extracted the tipset-class vectors to https://github.com/filecoin-project/oni/issues/220. So now this issue is only about message-class vectors, and it can be closed ;-)

PR #208, opened 2020-08-06
refactor: add default TestVector to TestDriver

alanshaw:
This PR adds a `TestVector` to the `TestDriver` so that test driver methods that apply messages can be recorded in the test vector.

refs #194

### DIFF
--- a/tvx/chain/address.go
+++ b/tvx/chain/address.go
@@ -44,7 +44,7 @@ func MustNewActorAddr(data string) addr.Address {
 	return address
 }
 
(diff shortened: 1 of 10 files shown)


nonsense:
Overall LGTM, but we need to fix up the compilation errors, as other commands are also using the schema that now goes into its own package.

raulk:
@alanshaw I see you're moving the Go schema. Please rebase on top of https://github.com/filecoin-project/oni/pull/209, which modified `schema.go` ;-)

PR #209, opened 2020-08-06
schema: add message receipts + `description` and `comments` informational fields

raulk:
# TODO

- [x] Adapt go entities.


corpus/single_message.json
```diff
@@ -1,29 +1,35 @@
...
-
+  "apply_messages": [
+    {
+      "msg": "0x89004300ba20583103b85358cc1c968d826c0a0efcde99b6e4db138edfc051f188a4471cfc23ccf7884adbbac912e45f7363774bf16f83ca8f1902b44800505e1442e66b744200011a09d225c80758c68219016f58c099a447a8c80988e1c061dd3a8dd674033c3a6083c4b55cf035e6e8de3baf09436ad00adf6eb14d58ae6691d22ce82255874047e7aa0b9a86bbf424afc6d89dc29f7b253dd04c88a625b86345f05aa0bdee503971c1e9735ed5b81c31ad41a1f10b387d96151d1cbd0c4286267daf04087fd8d5272f870994c3732f3fc4f54287a485d6c8dafb6056df3fc9233f849027b4c1272850435126fef65a8363768a858c0cc5f9d9114ff378ad50cab25ab0a2f685ca63d9f1f7727da4d06db5b80b5f",
+      "receipt": {
```
willscott, line 26: this seems like a validation condition rather than a thing we're applying.
Probably only relevant for messages, not for other classes.
maybe add to post conditions when in `messages` class vector, and expect the length of receipts to equal the length of messages being applied.

raulk: Good point; fixed in [`bdb3871` (#209)](https://github.com/filecoin-project/oni/pull/209/commits/bdb387120e40a0ae9e5b2f10d241f5da47fd90c4). Although JSON schema won't allow me to enforce:

- that `len(postconditions.receipts) == len(apply_messages)`.
- that when `class == messages`, `postconditions.receipts` must be present (although this is implicit in the above constraint)

Damn you, JSON Schema. I documented these constraints as descriptions; we'll need to enforce them in code.

PR #210, opened 2020-08-06
payment channel suite

nonsense:
This PR is moving the `payment channel suite` from https://github.com/filecoin-project/chain-validation/blob/master/suites/message/paych.go so that we build test vectors out of it.

In the process I am changing the schema for `ApplyMessage` to include an `Epoch`, so that we can simulate messages pushed to the VM that arrive in different epochs - necessary for one of the payment channel tests (when `settling`).


tvx/dummy_testingt.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import (
+	"fmt"
+	"os"
+)
+
+var t dummyTestingT
```
willscott, line 8: collapse back to the one in drivers?

nonsense: This tests do use `require` and `asserts` in the `main` package :/ If they are not too many, I will convert them.

tvx/suite_messages_paych.go
```diff
@@ -0,0 +1,229 @@
...
+		return err
+	}
+
+	err = func(testname string) error {
+		td := drivers.NewTestDriver()
+
+		v := newEmptyMessageVector()
+
+		//const pcTimeLock = abi_spec.ChainEpoch(1)
+		const pcTimeLock = abi_spec.ChainEpoch(0)
```
nonsense, line 76: This is a difference between our tests and `chain-validation` - it seems like they are using epoch 1 as an initial epoch for tests, whereas we are using 0.

tvx/suite_messages_paych.go
```diff
@@ -0,0 +1,229 @@
...
+
+		msg = td.MessageProducer.PaychSettle(receiver, paychAddr, nil, chain.Value(big_spec.Zero()), chain.Nonce(0))
+		v.ApplyMessages = append(v.ApplyMessages, Message{Bytes: chain.MustSerialize(msg)})
+		settleResult := td.ApplyOk(msg)
+
+		// advance the epoch so the funds may be redeemed.
+		td.ExeCtx.Epoch += paych_spec.SettleDelay
+
+		msg = td.MessageProducer.PaychCollect(receiver, paychAddr, nil, chain.Nonce(1), chain.Value(big_spec.Zero()))
+		v.ApplyMessages = append(v.ApplyMessages, Message{Epoch: &td.ExeCtx.Epoch, Bytes: chain.MustSerialize(msg)})
```
nonsense, line 200: The `collect` message should be set `paych_spec.SettleDelay` epochs later than the `settle` I think.

tvx/dummy_testingt.go
nonsense: They are not too many, but probably best to do that in one go after we transition the tests... no point converting `assertEqual` in one file, but keeping it in 10 others.

tvx/suite_messages_paych.go
```diff
@@ -0,0 +1,228 @@
...
+
+	"github.com/filecoin-project/oni/tvx/chain"
+	"github.com/filecoin-project/oni/tvx/drivers"
+)
+
+func MessageTest_Paych() error {
+	var initialBal = abi_spec.NewTokenAmount(200_000_000_000)
+	var toSend = abi_spec.NewTokenAmount(10_000)
+
+	err := func(testname string) error {
```
willscott, line 21: can these tests be factored out to top level functions?

nonsense: Yes, I imagine this will be part of the refactor @raulk mentioned.

tvx/dummy_testingt.go
nonsense: Falling back to drivers.T, although it is similarly ugly... :/

PR #212, opened 2020-08-07
feat: test driver automatically records applied messages in test vector

alanshaw:
The `TestDriver` now records applied messages in it's `TestVector` whenever an `Apply*` method is called. This removes the need for manual message serialization and recording in the tests.

### DIFF
--- a/tvx/drivers/test_driver.go
+++ b/tvx/drivers/test_driver.go
@@ -325,10 +325,11 @@ func (td *TestDriver) applyMessage(msg *types.Message) (result vtypes.ApplyMessa
 		}
 	}()
 
-	result, err := td.applier.ApplyMessage(td.ExeCtx.Epoch, msg)
+	epoch := td.ExeCtx.Epoch
+	result, err := td.applier.ApplyMessage(epoch, msg)
 	td.Vector.ApplyMessages = append(
 		td.Vector.ApplyMessages,
-		schema.Message{Epoch: &td.ExeCtx.Epoch, Bytes: chain.MustSerialize(msg)},
+		schema.Message{Epoch: &epoch, Bytes: chain.MustSerialize(msg)},
 	)
 	require.NoError(T, err)
 	return result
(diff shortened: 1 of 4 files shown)


tvx/suite_messages_paych.go
```diff
@@ -172,22 +167,17 @@ func MessageTest_Paych() error {
...
 
 		msg = td.MessageProducer.PaychSettle(receiver, paychAddr, nil, chain.Value(big_spec.Zero()), chain.Nonce(0))
-		td.Vector.ApplyMessages = append(td.Vector.ApplyMessages, schema.Message{Bytes: chain.MustSerialize(msg)})
 		settleResult := td.ApplyOk(msg)
 
 		// advance the epoch so the funds may be redeemed.
 		td.ExeCtx.Epoch += paych_spec.SettleDelay
 
 		msg = td.MessageProducer.PaychCollect(receiver, paychAddr, nil, chain.Nonce(1), chain.Value(big_spec.Zero()))
-		td.Vector.ApplyMessages = append(td.Vector.ApplyMessages, schema.Message{Epoch: &td.ExeCtx.Epoch, Bytes: chain.MustSerialize(msg)})
```
nonsense, line 189: @alanshaw note that this requires a specific epoch.

tvx/drivers/test_driver.go
```diff
@@ -326,8 +326,11 @@ func (td *TestDriver) applyMessage(msg *types.Message) (result vtypes.ApplyMessa
 	}()
 
 	result, err := td.applier.ApplyMessage(td.ExeCtx.Epoch, msg)
+	td.Vector.ApplyMessages = append(
+		td.Vector.ApplyMessages,
+		schema.Message{Bytes: chain.MustSerialize(msg)},
```
nonsense, line 331: We need to pass td.ExeCtx.Epoch here as well, as part of `schema.Message`.

tvx/drivers/test_driver.go
```diff
@@ -374,8 +377,11 @@ func (td *TestDriver) applyMessageSigned(msg *types.Message) (result vtypes.Appl
 		Signature: msgSig,
 	}
 	result, err = td.applier.ApplySignedMessage(td.ExeCtx.Epoch, smsgs)
+	td.Vector.ApplyMessages = append(
+		td.Vector.ApplyMessages,
+		schema.Message{Bytes: chain.MustSerialize(smsgs)},
```
nonsense, line 382: We need to pass `td.ExeCtx.Epoch` here as well, as part of `schema.Message`.

PR #214, opened 2020-08-07
feat: add test name to test vector meta

alanshaw:
So we can more easily distinguish between test vectors.


tvx/suite_messages_message_application.go
```diff
@@ -264,6 +270,7 @@ func MessageTest_MessageApplicationEdgecases() error {
 
 	err = func(testname string) error {
 		td := drivers.NewTestDriver()
+		td.Vector.Meta.Desc = testname
```
nonsense, line 273:
```suggestion
		td.Vector.Meta.Desc = testname
		td.Vector.Meta.Comment = "Note that this test is not a valid message, since it is using an unknown actor. However in the event that an invalid message isn't filtered by block validation we need to ensure behaviour is consistent across VM implementations."
```

PR #215, opened 2020-08-07
messages/nested suite

nonsense:
This PR is moving the `messages/nested` suite to Project Oni, and generating test vectors for other impl. to consume based on it.

TODO:
- [x] figure out what's up with this puppet actor...

### DIFF
--- a/tvx/exec_lotus.go
+++ b/tvx/exec_lotus.go
@@ -75,7 +75,7 @@ func runExecLotus(_ *cli.Context) error {
 }
 
 func executeTestVector(tv schema.TestVector) error {
-	fmt.Println("executing test vector")
+	fmt.Println("executing test vector:", tv.Meta.Desc)
 	switch tv.Class {
 	case "message":
 		var (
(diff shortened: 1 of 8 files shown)


tvx/exec_lotus.go
```diff
@@ -75,7 +75,7 @@ func runExecLotus(_ *cli.Context) error {
 }
 
 func executeTestVector(tv schema.TestVector) error {
-	fmt.Println("executing test vector")
+	fmt.Println("executing test vector:", tv.Meta.Desc)
```
alanshaw, line 78: 👍

tvx/lotus/driver.go
```diff
@@ -42,6 +43,13 @@ func (d *Driver) ExecuteMessage(msg *types.Message, preroot cid.Cid, bs blocksto
 	if err != nil {
 		return nil, cid.Undef, err
 	}
+	// need to modify the VM invoker to add the puppet actor
```
nonsense, line 46: Adding the puppet actor to the VM, so that we can test messages that would otherwise fail preconditions.

PR #227, opened 2020-08-13
state tree examiner

willscott:
Standalone tool for visualization of diff's between state trees.

Two subcommands:
* vector - takes in a json test vector and presents a visualization of what changes in the pre and post stateroots of the car
* diff - takes in a CID object (message or block) or pair (block1..block2) and presents a visualization of the change over them.

By default, expands to the actor level (balances, which actors change), as full expansion/retrieval is quite slow. Full actor expansion is enabled with `--expand-actors all`. Specific actors can be expanded with `--expand-actors reward,cron` or `--expand-actors bafkqaetgnfwc6mjpon2g64tbm5sw22lomvza` (either the class of the builtin actor, the code Cids, or actor head Cids)


fc-examine/schema.go
```diff
@@ -0,0 +1 @@
+../tvx/schema.go
```
raulk, line 1: is this some Go magic trick I don't know about? 😮

willscott: a git symbolic link :)

fc-examine/go.mod
```diff
@@ -0,0 +1,26 @@
...
+	github.com/multiformats/go-multiaddr v0.2.2
+	github.com/multiformats/go-multiaddr-net v0.1.5
+	github.com/urfave/cli/v2 v2.2.0
+	github.com/whyrusleeping/cbor-gen v0.0.0-20200812213548-958ddffe352c
+	github.com/willscott/go-cmp v0.5.2-0.20200812183318-8affb9542345
+)
+
+replace github.com/filecoin-project/filecoin-ffi => ../extra/filecoin-ffi
+
+replace github.com/filecoin-project/sector-storage => github.com/filecoin-project/lotus/extern/sector-storage v0.0.0-20200814115354-002e972f7535
```
raulk, line 26: After https://github.com/filecoin-project/lotus/pull/3089/, we should no longer require this.

fc-examine/lib/examine.go
```diff
@@ -0,0 +1,450 @@
...
+)
+
+type config struct {
+	ExpandActors   bool
+	ActorCidFilter []cid.Cid
+}
+
+type Option func(c *config)
+
+func ExpandActors(c *config) {
```
raulk, line 43: Godocs, please?

fc-examine/lib/examine.go
```diff
@@ -0,0 +1,450 @@
...
+	c.ExpandActors = true
+}
+
+func ExpandActorByCid(cids []cid.Cid) Option {
```
raulk, line 47: Godocs, please?

fc-examine/lib/examine.go
raulk: An example of the same diff with and without this option would be 🔝

fc-examine/lib/examine.go
```diff
@@ -0,0 +1,450 @@
...
+	return func(c *config) {
+		c.ExpandActors = true
+		c.ActorCidFilter = cids
+	}
+}
+
+// Parse a user entered fuzzy definition for actor expansion.
```
raulk, line 54: I have no clue what this sentence means :-)

raulk: Examples of well-formed arguments would be welcome.

fc-examine/lib/examine.go
```diff
@@ -0,0 +1,450 @@
...
+				k = nk
+			}
+			m[k] = &template
+
+			return nil
+		})
+		return m
+	}
+
+	initHampTransformer := func(n *hamtNode) map[string]string {
```
raulk, line 144: hamp~p~t? Or is there something I'm missing?

fc-examine/lib/examine.go
```diff
@@ -0,0 +1,450 @@
...
+	cidMap[".*\\(miner\\.State\\)\\.PreCommittedSectorsExpiry$"] = reflect.TypeOf("")
+	cidMap[".*\\(miner\\.State\\)\\.AllocatedSectors$"] = reflect.TypeOf((*bitfield.BitField)(nil))
+	cidMap[".*\\(miner\\.State\\)\\.Sectors$"] = reflect.TypeOf((*adt.Array)(nil))
+	cidMap[".*\\(miner\\.State\\)\\.Sectors.*SealedCID$"] = reflect.TypeOf("")
+	cidMap[".*\\(miner\\.State\\)\\.Deadlines$"] = reflect.TypeOf((*storageMinerActor.Deadlines)(nil))
+	cidMap[".*\\(miner\\.State\\)\\.Deadlines.*Due([^\\.]*)$"] = reflect.TypeOf((*storageMinerActor.Deadline)(nil))
+	cidMap[".*\\(miner\\.State\\).*Partitions$"] = reflect.TypeOf(make([]*storageMinerActor.Partition, 0))
+	cidMap["miner\\.Deadline\\)\\.ExpirationsEpochs$"] = reflect.TypeOf(make([]*bitfield.BitField, 0))
+	cidMap[".*\\(miner\\.State\\).*Partitions.*ExpirationsEpochs$"] = reflect.TypeOf(make([]*storageMinerActor.ExpirationSet, 0))
+	cidMap[".*\\(miner\\.State\\).*EarlyTerminated$"] = reflect.TypeOf(make([]*bitfield.BitField, 0))
```
raulk, line 234: Can we use backticks to avoid the escaping within the regex?

fc-examine/lib/examine.go
```diff
@@ -0,0 +1,450 @@
...
+		cidMap[".*\\.Head$"] = reflect.TypeOf("")
+	}
+	cmpOpts = append(cmpOpts, cidTransformer(ctx, store, cborStore, cidMap)...)
+	if d := cmp.Diff(a, b, cmpOpts...); d != "" {
+		fmt.Printf("Diff: %v\n", d)
+	}
+}
+
+func cidTransformer(ctx context.Context, store blockstore.Blockstore, cborStore cbor.IpldStore, atlas map[string]reflect.Type) []cmp.Option {
+	options := make([]cmp.Option, 0)
```
raulk, line 277:
```suggestion
	var options []cmp.Option
```

raulk (changes requested):
I can't claim I understand everything that's going on here ;-) I'm gonna take it out for a spin now and will report more back.

One proposal I'd like to make is to follow the standard golang packaging, as far as possible. I would move the contents of `lib` to the root, and move the command to a main package in `cmd/statediff`.

This makes it easy to import as (in the future):

```go
import "github.com/filecoin-project/lotus/statediff"
```

Instead of the more awkward import (that also requires an import alias):

```go
import statediff "github.com/filecoin-project/lotus/statediff/lib"
```

It also paves the way for integration in Lotus, where you'll want the command to live inside `lotus/cmd/statediff`, and call out to `github.com/filecoin-project/lotus/statediff`.

fc-examine/lib/examine.go
```diff
@@ -0,0 +1,450 @@
...
+	"storagePower":     builtin.StoragePowerActorCodeID,
+	"storageMiner":     builtin.StorageMinerActorCodeID,
+	"storageMarket":    builtin.StorageMarketActorCodeID,
+	"paymentChannel":   builtin.PaymentChannelActorCodeID,
+	"multisig":         builtin.MultisigActorCodeID,
+	"reward":           builtin.RewardActorCodeID,
+	"verifiedRegistry": builtin.VerifiedRegistryActorCodeID,
+}
+
+func Diff(ctx context.Context, store blockstore.Blockstore, a, b cid.Cid, opts ...Option) {
```
raulk, line 92: Having this print straight to stdout is a bit inconvenient. Can we make it return a string? Or provide an alternate `Sdiff` that returns a `string`?

willscott:
ported to github.com/filecoin-project/statediff

PR #229, opened 2020-08-13
test vector builder API + partial migration of suites

raulk:
This PR introduces a new flavour of the builder API that's a bit more semantically structured and tightly integrated with test vector production.

It lives in package `builders` for now, and the central object is `builders.Builder`. There will be one entrypoint/constructor per test vector class. Right now, there's `builders.MessageVector(metadata)`.

Building a message vector has three stages:

1. preconditions.
2. applies.
3. checks.

In the precondition stage, you initialize accounts, actors, etc.
In the applies stage, you accumulate the messages that will be applied.
In the checks stage, you perform a sanity check.

Calling `v.CommitPreconditions()` finalizes the precondition stage by taking a snapshot of the precondition state tree. The applies stage commences.

Calling `v.CommitApplies()` finalizes the applies stage by applying all messages and capturing the postconditions (state tree and receipts). The checks stage commences.

Calling `v.Finished(io.Writer)` declares this test vector as succeeded, and spits out the test vector JSON to the provided `io.Writer` (usually `os.Stdout`).

All throughout, you can perform assertions using the `v.Assert` object. It embeds the full [testify toolkit](https://github.com/stretchr/testify), and enhances it with domain-specific asserters.

---

There are cases in the applies stage where certain the input of certain messages depends on the output of previous messages. For those cases, you can apply messages individually using `v.Messages.ApplyOne(*ApplicableMessage)` or `v.Messages.ApplyN(*ApplicableMessage...)`. The following rules apply when using individual applications:

1. Messages must be applied in the order they are enrolled. Out-of-order applications will fail.
2. `CommitApplies()` will apply any trailing unapplied messages.

---

This PR also migrates all suites to this new API.

---

# TODO

- [x] individual applies.
- [ ] migrate suite_messages_create_actor.go
- [x] migrate suite_messages_message_application.go
- [ ] migrate suite_messages_multisig.go
- [ ] migrate suite_messages_nested.go
- [X] migrate suite_messages_paych.go
- [ ] migrate suite_messages_transfer.go


tvx/scripts/paych2.go
```diff
@@ -0,0 +1,191 @@
...
+
+	expectedActorAddr := AddressHandle{
+		ID:     MustNewIDAddr(MustIDFromAddress(receiver.ID) + 1),
+		Robust: sender.NextActorAddress(0, 0),
+	}
+
+	// Verify init actor return.
+	var ret init_.ExecReturn
+	MustDeserialize(createMsg.Result.Return, &ret)
+	v.Assert.Equal(expectedActorAddr.Robust, ret.RobustAddress)
```
nonsense, line 52: Nitpick: I am not sure why the asserters are part of the test vector. I think these are ortogonal and don't belong together.

tvx/builders/messages.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	msg := &types.Message{
+		To:       to,
+		From:     from,
+		Nonce:    options.nonce,
+		Value:    options.value,
+		Method:   method,
+		Params:   params,
+		GasPrice: options.gasPrice,
```
nonsense, line 59: Note that this is already outdated. We now have `GasFeeCap` and `GasPremium`.

tvx/scripts/paych2.go
raulk: To (a) enable us to contextualise the error message and (b) so that it has access to the builder's context, including the state tree, messages, etc.

tvx/builders/messages.go
raulk: Yup. Gotta do an upgrade and deal with the breakage.

nonsense:
LGTM

Issue #189, opened 2020-08-05
batch runner for test vector generation scripts


raulk:
***Let's write this in go, as a subcommand of tvx***

We expect developers to use the builder DSL to assemble test vectors, by writing small go scripts with the DSL being ported over from chain validation in https://github.com/filecoin-project/oni/pull/181.

Imagine the following layout:

```
gen
|_ paych_suite
|   |__ error_cases.go
|   |__ happy_paths.go
|   |__ mario.go
|   |__ luigi.go
|
|_ init_suite
|   |__ zelda.go
|   |__ peach.go
|
|_ toad_suite
    |__ bowser.go
```

We need a script that accepts `gen`, or `paych_suite`, or `init_suite` as an argument, and runs all go scripts within it using `go run <path>`. A second argument indicates the output directory.

Each script may generate _many_ test vectors, delimited by newlines, i.e. using ndjson (one test vector per line). We should save each line to a json file with name: `<go_file_name>_<seq>.json`.

The script would replicate the same directory layout. So running:

```sh
$ batch-gen --in-dir <path_to_gen> --out-dir <target_dir>
```

Would generate:

```
target
|_ paych_suite
|   |__ error_cases_0.json   (assuming gen/paych_suite/error_cases.go generates 4 lines)
|   |__ error_cases_1.json
|   |__ error_cases_2.json
|   |__ error_cases_3.json
|   |__ happy_paths_0.json   (assuming gen/paych_suite/happy_paths.go generates 3 lines)
|   |__ happy_paths_1.json
|   |__ happy_paths_2.json
|   |__ mario_0.json         (assuming gen/paych_suite/mario.go generates 2 lines)
|   |__ mario_1.json
|   |__ luigi_0.json         (assuming gen/paych_suite/luigi.go generates 2 lines)
|   |__ luigi_1.json
|
|_ init_suite
|   |__ zelda_0.json         (assuming gen/init_suite/zelda.go generates 3 lines)
|   |__ zelda_1.json
|   |__ zelda_2.json
|   |__ peach_0.json         (assuming gen/init_suite/peach.go generates 1 line)
|
|_ ...
```

raulk:
@nonsense I think you're the best person to tackle this one this week, maybe pairing with @alanshaw; wdyt?

PR #232, opened 2020-08-14
generator; single main() entrypoint for each suite.

raulk:
This introduces a new organisation for test vector scripts.

Each suite now lives in a package (directory), with a single main entrypoint.

I have introduced a `Generator` that the main method can delegate to to have it handle the generation and writing of test vectors into properly named files. I have also taken the opportunity to make things DRYer; scripts no longer have to instantiate the `Builder` manually, or call `v.Finish()`: the `Generator` takes care of that.

See godocs of the `Generator` struct for more info on the layout.

I have migrated the `message_application` suite to use this style. `paych` suite pending.

---

This fixes https://github.com/filecoin-project/oni/issues/189 via an alternative approach that is less hacky and a lot more declarative, and DRY.

---

Try it out:

```shell script
$ cd tvx
$ go run ./scripts/msg_application -o ./foo
generating test vector: msg-apply-fail-actor-execution-illegal-arg
generating test vector: msg-apply-fail-unknown-receiver
generating test vector: msg-apply-fail-unknown-receiver
generating test vector: msg-apply-fail-invalid-receiver-method
generating test vector: msg-apply-fail-invalid-receiver-method
generating test vector: msg-apply-fail-onchainsize-gas
generating test vector: msg-apply-fail-transfer-accountcreation-gas
generating test vector: msg-apply-fail-transfer-accountcreation-gas
$ ls -lrt foo
total 208
-rw-r--r--  1 raul  staff   2720 Aug 14 15:59 invalid_msgs-0.json
-rw-r--r--  1 raul  staff   2720 Aug 14 15:59 invalid_msgs-1.json
-rw-r--r--  1 raul  staff   3142 Aug 14 15:59 unknown_actors-1.json
-rw-r--r--  1 raul  staff   3142 Aug 14 15:59 unknown_actors-0.json
-rw-r--r--  1 raul  staff   4045 Aug 14 15:59 actor_exec-0.json
-rw-r--r--  1 raul  staff  26436 Aug 14 15:59 gas_cost-2.json
-rw-r--r--  1 raul  staff  26436 Aug 14 15:59 gas_cost-1.json
-rw-r--r--  1 raul  staff  26436 Aug 14 15:59 gas_cost-0.json
```


tvx/builders/generator.go
```diff
@@ -0,0 +1,166 @@
...
+	)
+
+	flag.Parse()
+
+	ret := new(Generator)
+
+	// Output directory is compulsory.
+	if *outputDir == "" {
+		fmt.Printf("output directory not specified\n")
+		os.Exit(1)
```
nonsense, line 62:
```suggestion
		log.Fatalf("output directory not specified\n")
```

tvx/builders/generator.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	err := ensureDirectory(*outputDir)
+	if err != nil {
+		log.Println(err)
+		os.Exit(1)
```
nonsense, line 68:
```suggestion
		log.Fatal(err)
```

tvx/builders/generator.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	ret.OutputPath = *outputDir
+
+	// If a filter has been provided, compile it into a regex.
+	if *filter != "" {
+		exp, err := regexp.Compile(*filter)
+		if err != nil {
+			fmt.Printf("supplied regex %s is invalid: %s\n", *filter, err)
+			os.Exit(1)
```
nonsense, line 78:
```suggestion
			log.Fatalf("supplied regex %s is invalid: %s\n", *filter, err)
```

tvx/builders/generator.go
```diff
@@ -0,0 +1,166 @@
...
+			out, err := os.OpenFile(file, os.O_WRONLY|os.O_CREATE, 0644)
+			if err != nil {
+				fmt.Printf("failed to write to file %s: %s\n", file, err)
+				return
+			}
+			wg.Add(1)
+			go func(item *MessageVectorGenItem) {
+				g.generateOne(out, item)
+				wg.Done()
+			}(&item)
```
nonsense, line 112: Is that right? It is not obvious what the type of `item` is, so maybe just `item := item` to scope it?

tvx/builders/generator.go
```diff
@@ -0,0 +1,166 @@
...
+//  -o <directory>
+//		directory where test vector JSON files will be saved.
+//
+//  -f <regex>
+//		regex filter to select a subset of vectors to execute; matched against
+//	 	the vector's ID.
+//
+// Scripts can bundle test vectors into "groups". The generator will execute
+// each group in parallel, and will write each vector in a file:
+// <output_dir>/<group>-<sequential>.json
```
alanshaw, line 35:
```suggestion
// <output_dir>/<group>-<vector_ID>.json
```

Might be more stable and not cause problems when re-generating vectors over existing files (easier to identify also).

tvx/builders/generator.go
```diff
@@ -0,0 +1,166 @@
...
+		outputDir = flag.String("o", "", "output directory")
+		filter    = flag.String("f", "", "filter regex for test vectors to generate")
+	)
+
+	flag.Parse()
+
+	ret := new(Generator)
+
+	// Output directory is compulsory.
+	if *outputDir == "" {
```
alanshaw, line 60: Hmm, I thought the idea was to output to `os.Stdout` if this was not set?

raulk: Yup! I'm gonna implement that now. Was thinking on the spot during the call ;-)

alanshaw: I like your style 😁

tvx/builders/generator.go
raulk: Done!

tvx/builders/generator.go
raulk: Implemented.

tvx/builders/generator.go
raulk: Implemented.

tvx/builders/generator.go
raulk: Implemented.

Context: we were suppressing logs in the `init` function of `builder.go`, to not muddy the output JSON, that's why I was using `fmt.Print*` directly. But @willscott suggested to send the log package to `os.Stderr`, which keeps Unix pipes working, but it muddies the console output. To make things better, I have prefix `log` output with `>>>`.

tvx/builders/generator.go
raulk: Fixed.

Issue #183, opened 2020-08-05
switch test vector state binary fields to gzip + base64; try brotli


raulk:
This allows for shorter test vectors. We could also try Brotli instead of gzip, should give us a better compression ratio.

Tasks:

- [x] Switch to gzip (after: gzip + hex).
- [x] Switch to base64 (after: gzip + base64).
- [ ] Try Brotli (after: Brotli + base64).

Re: Brotli, I was inspired by https://dropbox.tech/infrastructure/-broccoli--syncing-faster-by-syncing-less.

alanshaw:
In our colo call today we decided the size reduction with [brotli+base64](https://github.com/filecoin-project/oni/pull/234) was not worth the additional UX/DX for an additional saving of ~3% over [gzip+base64](https://github.com/filecoin-project/oni/pull/233).

PR #234, opened 2020-08-14
refactor: switch to brotli+base64 encoding for binary data

alanshaw:
~32% smaller vectors:

```console
$ ./tvx suite-messages > vectors.brotli.b64.ndjson
$ ls -lah vectors.*
-rw-r--r--  1 alan  staff   119K 14 Aug 16:07 vectors.gzip.brotli.ndjson
-rw-r--r--  1 alan  staff   174K 14 Aug 11:31 vectors.gzip.hex.ndjson
```

Note, only ~3% smaller than [gzip+base64](https://github.com/filecoin-project/oni/pull/233).

refs #183


alanshaw:
N.B. There is no pure _go_ brotli writer (I found a project with a reader but no writer yet) so here we're using google's `cbrotli` which has additional tooling and library requirements.

raulk:
Thanks for the experiment! In the call today, we decided to go with b64+gzip and discard the Brotli idea, since it introduces build time complexity due to the lack of a Golang-native Brotli implementation, and the little gain we get compared to gzip.

PR #238, opened 2020-08-15
upgrade test vectors to latest lotus

raulk:
* Pulls in the Dynamic Network Fee changes.
* Adds new assertion methods `MessageSendersSatisfy` and `EveryMessageSenderSatisfies`, with new predicate type `ActorPredicate` with out-of-the-box functions `BalanceUpdated` and `NonceUpdated`.
* Performs assertions equivalent to chain-validation's `AssertActorState` in the migrated vectors. 
* Remove chain-validation from our source tree 👋 
* Underscore unmigrated suites to prevent compilation.


raulk:
This is passing and looks good. I'm going to merge so I can start referencing from other places. We can discuss on Monday if something looks off.

PR #239, opened 2020-08-16
migrate actor creation suite.

raulk:


raulk:
Merging in order to make progress and publish a v0 corpus.

PR #240, opened 2020-08-16
migrate transfer suite.

raulk:
Original version: https://github.com/filecoin-project/chain-validation/blob/4ea86efa4470971ee8111f6f96b7ba674d4b3cc8/suites/message/transfer.go

Vectors materialized: https://github.com/filecoin-project/test-vectors/pull/1/commits/ae1d3e876b6e614bbe9a3991a221518119490a1d

Notes:

- I did not migrate the "ok transfer from known address to new account" case because it's redundant with the actor creation cases.


raulk:
Merged to make progress with corpus v0.

PR #245, opened 2020-08-28
payment channels tests improvements

dirkmc:
Improve payment channels tests:
- Send vouchers from channel sender to channel receiver
- Channel receiver adds vouchers to local datastore
- Channel receiver checks that vouchers are spendable
- Channel sender invokes Submit message on chain
- Channel receiver checks that vouchers are no longer spendable


lotus-soup/paych/stress.go
```diff
@@ -59,102 +74,236 @@ func Stress(t *testkit.TestEnvironment) error {
...
+
+	case ModeSender:
+		err := runSender(ctx, t, clients, cl)
 		if err != nil {
-			return fmt.Errorf("failed to acquire wallet balance: %w", err)
+			return err
 		}
+	}
+
+	t.SyncClient.MustSignalEntry(ctx, testkit.StateDone)
```
dirkmc, line 90: @raulk for some reason the bootstrapper never exits, even though I'm signalling StateDone here

raulk: Right! I think this is because the miners have not signalled `StateDone` (the boostrapper waits for all instances to signal).

The shutdown procedure is:

- clients are done, and signal on `StateDone`, and also on `StateStopMining` to inform miners that they can stop mining now.
- miners wait for all clients to signal on `StateStopMining`, and only then do they signal on `StateDone`.
- by now, all nodes should've signalled on `StateDone`, and the bootstrapper will shut down.

raulk (approved):
Thanks for getting all of this in shape!!! Here's some energy replenishment for you:

raulk:
@dirkmc would it be useful to integrate this test plan into Lotus CI?

PR #249, opened 2020-09-09
upgrade lotus-soup (drand, lotus, stop/resume drand)

nonsense:
This PR is an update to `lotus-soup` test while running the manual Drand outage tests. It is:

1. Upgrading Lotus
2. Fixing start/stopping of Drand
3. Upgrading Drand library
4. Upgrading proof parameters
5. Bumping up timeouts on Drand side, so that preparation of nodes doesn't fail with real production parameters.

---

FOLLOW UP PR:
- [ ] re-enable payment channel tests
- [ ] upgrade to latest Lotus v0.7.x


lotus-soup/testkit/role_drand.go
```diff
@@ -127,16 +128,18 @@ func (dr *DrandInstance) RunDKG(nodes, thr int, timeout string, leader bool, lea
 
 func (dr *DrandInstance) Halt() {
 	dr.t.RecordMessage("drand node #%d halting", dr.t.GroupSeq)
-	dr.daemon.StopBeacon()
```
nonsense, line 130: This is a major change in this PR. Using `StopBeacon` to simulate Drand outage is probably not desirable, as it doesn't really test any possible real world scenario - we are stopping the beacon, but GRPC connections stay open, which essentially results in blocked requests once Drand is resumed.

Therefore I am changing to `Stop` which also closes any connections and stops HTTP APIs.

.circleci/config.yml
```diff
@@ -41,17 +41,9 @@ workflows:
   version: 2
   main:
     jobs:
-      - tvx-build-test-linux
       - soup-build-linux
 
 jobs:
-  tvx-build-test-linux:
```
nonsense, line 48: I believe we no longer need this job in CI.

lotus-soup/manifest.toml
```diff
@@ -10,8 +10,8 @@ enabled = true
 
 [builders."docker:go"]
 enabled = true
-build_base_image = "iptestground/oni-buildbase:v5"
-runtime_image = "iptestground/oni-runtime:v3-debug"
+build_base_image = "iptestground/oni-buildbase:v6"
+runtime_image = "iptestground/oni-runtime:v5-debug-anton"
```
willscott, line 14: intentional?

lotus-soup/manifest.toml
```diff
@@ -90,6 +90,37 @@ instances = { min = 1, max = 100, default = 5 }
...
+  # Params relevant to drand nodes. drand nodes should have role="drand", and must all be
+  # in the same composition group. There must be at least threshold drand nodes.
+  # To get lotus nodes to actually use the drand nodes, you must set random_beacon_type="local-drand"
+  # for the lotus node groups.
+  drand_period = { type = "duration", default="30s" }
+  drand_catchup_period = { type = "duration", default="10s" }
+  drand_threshold = { type = "int", default = 2 }
+  drand_gossip_relay = { type = "bool", default = true }
+  drand_log_level = { type = "string", default="info" }
+  suspend_events = { type = "string", default="", desc = "a sequence of halt/resume/wait events separated by '->'" }
```
willscott, line 117: can it be directly specified as a string array in toml rather than 2 parsing formats?

willscott (approved):

lotus-soup/manifest.toml
nonsense: No, I will fix it.

lotus-soup/manifest.toml
nonsense: Yeah, it could be, but I don't think it matters much at this stage.

Honestly I am not sure that I like the halt/resume sequence to be time based and not epoch based as a start (time in Lotus is almost always expressed as epochs).

In any case the issues with these tests in general (automation, maintain dependencies, etc.) are much bigger I think.

PR #252, opened 2020-09-10
wip graphsync stress test.

raulk:
A stress test that takes three parameters:

* size: the filesize to test with (default: 100MB).
* latencies: an array of latencies to set on the provider side.
* bandwidths: an array of bandwidths (egress per second) to set on the provider side.

It supports two groups:

* `providers`, max instance 1.
* `requestors`, as many as needed, right now testing with 1.

Both provider and requestor will do an iteration for every latency x bandwidth tuple.

For every iteration, the provider generates a new random file, and publishes the CID to a sync topic. It then sets the network latency for that round. The requestor awaits the CID, and also awaits for the network to be configured, and then performs a graphsync request for the CID.

## Instructions

To run, **first install testground as indicated [in the Testground README](https://github.com/testground/testground#getting-started).**

Then clone this repo and this branch:

```shell script
$ testground daemon # start the testground daemon in one tab

$ git clone https://github.com/filecoin-project/oni.git
$ cd oni
$ testground plan import --from graphsync
created symlink /Users/raul/testground/plans/graphsync -> graphsync
imported plans:
graphsync stress
$ testground run composition -f ./graphsync/_compositions/stress.toml
...
```

## TODO

All readings seem to be the same. This is odd. I think graphsync works asynchronously, but I'm not waiting for each request to complete. @hannahhoward?


graphsync/main.go
```diff
@@ -0,0 +1,340 @@
...
+
+	host, peers := makeHost(ctx, runenv, initCtx)
+	defer host.Close()
+
+	var (
+		// make datastore, blockstore, dag service, graphsync
+		ds     = dss.MutexWrap(datastore.NewMapDatastore())
+		bs     = blockstore.NewBlockstore(ds)
+		dagsrv = merkledag.NewDAGService(blockservice.New(bs, offline.Exchange(bs)))
+		gsync  = gsi.New(ctx, gsnet.NewFromLibp2pHost(host), makeLoader(bs), makeStorer(bs))
```
hannahhoward, line 88: you can get rid of the makeLoader and makeStorer functions and just use these utility functions here that do the exact same thing (I should have removed them from the integration tests to not create confusion)

https://github.com/ipfs/go-graphsync/blob/master/storeutil/storeutil.go#L16
https://github.com/ipfs/go-graphsync/blob/master/storeutil/storeutil.go#L32

graphsync/main.go
```diff
@@ -0,0 +1,340 @@
...
+
+	switch runenv.TestGroupID {
+	case "providers":
+		if runenv.TestGroupInstanceCount > 1 {
+			panic("test case only supports one provider")
+		}
+
+		runenv.RecordMessage("we are the provider")
+		defer runenv.RecordMessage("done provider")
+
```
hannahhoward, line 99: I would add the following here:
```suggestion
gsync.RegisterIncomingRequestHook(func(p peer.ID, request gs.RequestData, hookActions gs.IncomingRequestHookActions) {
hookActions.ValidateRequest()
})
```

Otherwise your requests will get rejected.

graphsync/main.go
```diff
@@ -0,0 +1,340 @@
...
+			<-initCtx.SyncClient.MustBarrier(ctx, stateNet, 1).C
+			runenv.RecordMessage("network configured for round %d", round)
+
+			// execute the traversal.
+			runenv.RecordMessage(">>>>> requesting")
+			progressCh, errCh := gsync.Request(ctx, p.ID, clink, sel)
+			for r := range progressCh {
+				runenv.RecordMessage("******* progress: %+v", r)
+			}
+
```
hannahhoward, line 162: One thing you can do here is reread out the file to make sure you have all the bytes -- though you'd need to pass the original file bytes through with the CID in the channel (that might be a bit inefficient -- or maybe the processes can can both read the same system file? I'm not sure the constraints in testground for sharing large chunks of data)

You can see the code you can use to verify the bytes on the other side here: https://github.com/ipfs/go-graphsync/blob/master/impl/graphsync_test.go#L847-L865

graphsync/main.go
```diff
@@ -0,0 +1,340 @@
...
+			runenv.RecordMessage("<<<<< request complete")
+			if len(errCh) > 0 {
```
hannahhoward, line 164: this is possibly poor design on graphsync's part, but this will not work as a way to determine if there are errors.
graphsync collects responses and errors in an array buffer so they can be read asynchronously by caller of request at any time. (https://github.com/ipfs/go-graphsync/blob/master/requestmanager/responsecollector.go) A side affect is the len() on a channel will always be zero.

you need to iterate over the error channel to determine if there are errors. Don't worry -- iterating over the response and error channels can be done independently without blocking the other. if you iterate over the error channel after already iterating on the progress channel, it will not block at all.

hannahhoward (changes requested):
Most of my comments are suggestions, but two will cause problems for you:
1. You need to add the validation to the responder
2. You are probably currently getting an error but missing it cause the check on the error channel is not an accurate way to read if there are errors

graphsync/main.go
raulk: Done, thanks!

PR #254, opened 2020-09-16
upgrade lotus to master, and sync filecoin-ffi versions

nonsense:
This PR is:
1. Upgrading lotus to master (actually to commit https://github.com/filecoin-project/lotus/pull/3862/commits/b4115021c5063ec5c12a44f3b4d62c055f79d8e5 so that we can compile)
2. Syncing `filecoin-ffi` with the version used on Lotus.
3. Adding a git submodule for `fil-blst`.

PR #256, opened 2020-09-16
clone fil-blst in extra for the local:docker runner

nonsense:
Follow up to https://github.com/filecoin-project/oni/pull/254

We need to have the `fil-blst` as part of the docker image `buildbase` when using the `local:docker` testground runner.

We also need to upload that folder as part of the extra sources when using the `local:exec` runner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ Additional to the Filecoin FFI Git submodules, we are also bundling `proof param
 * `v4` => locking in Filecoin FFI commit 6a143e06f923f3a4f544c7a652e8b4df420a3d28.
 * `v5` => locking in Filecoin FFI commit cddc56607e1d851ea6d09d49404bd7db70cb3c2e.
 * `v6` => locking in Filecoin FFI commit 40569104603407c999d6c9e4c3f1228cbd4d0e5c.
+* `v7` => add Filecoin-BLST repo to buildbase.
 
 
 ### oni-runtime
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -222,6 +222,7 @@ Additional to the Filecoin FFI Git submodules, we are also bundling `proof param
 * `v4` => locking in Filecoin FFI commit 6a143e06f923f3a4f544c7a652e8b4df420a3d28.
 * `v5` => locking in Filecoin FFI commit cddc56607e1d851ea6d09d49404bd7db70cb3c2e.
 * `v6` => locking in Filecoin FFI commit 40569104603407c999d6c9e4c3f1228cbd4d0e5c.
+* `v7` => add Filecoin-BLST repo to buildbase.
```
willscott, line 225: * should we track the BLST commit hash here in the same way as the FFI one?
* is this gonna imply that lotus in submoduling oni now needs to keep versions for both of these in sync?

willscott (approved):

README.md
nonsense: > should we track the BLST commit hash here in the same way as the FFI one?

I think so. I don't see a way around it.

> is this gonna imply that lotus in submoduling oni now needs to keep versions for both of these in sync?

Yes. However since Magik also requested that we use Lotus' version of FFI, I added replaces for all the dependencies. This means that when compiling `lotus-soup` in the `lotus` repo, we are always using the `lotus` dependencies and make sure it compiles (to me this sounds like less overhead to the Lotus team).

Obviously we would need to on-demand update versions in `oni/lotus-soup` itself (for example here)...

---

I don't like this, although I don't have any better ideas on how to improve right now.

PR #261, opened 2020-09-24
combine #260 and #246; remove driver (now in Lotus)

raulk:
Fixes #260.
Fixes #246.

---

Also, remove redundant driver from here.


tvx/examine.go
```diff
@@ -14,7 +14,8 @@ import (
 	init_ "github.com/filecoin-project/specs-actors/actors/builtin/init"
```
willscott, line 14: should we leave it as a 'todo' to migrate these specs-actors' references to the stable versions?

willscott (approved):

tvx/examine.go
raulk: Oh good point, this needs to be migrated!

nonsense (approved):
overall lgtm

raulk:
Merging this to close it out, but ultimately `tvx` is being from here and ported over to Lotus in https://github.com/filecoin-project/lotus/pull/4064.

PR #262, opened 2020-09-24
fix lotus-soup/deals-e2e test case

nonsense:


.circleci/config.yml
```diff
@@ -64,10 +64,10 @@ jobs:
       - setup
       - run:
           name: "download testground"
-          command: wget https://gist.github.com/nonsense/5fbf3167cac79945f658771aed32fc44/raw/2e7ced2b1d0ca54033be57299e3fefb840958dfb/testground-3ee0dae -O ~/testground-cli && chmod +x ~/testground-cli
+          command: wget https://gist.github.com/nonsense/5fbf3167cac79945f658771aed32fc44/raw/35cc63898502274087da843b4752fda84025b861/testground-0abaea1 -O ~/testground-cli && chmod +x ~/testground-cli
```
willscott, line 67: can we pull from a branch so that this config doesn't need to be updated with each change to testground cli build?

willscott (approved):

.circleci/config.yml
nonsense: I want to avoid having to build testground from scratch here. Also the client side rarely changes... the requests for a `build` and `run` tasks change rarely. They changed now as I needed to propagate a field for `CreatedBy` (i.e. no authentication).

In any case I will make this more formal so that folks now how to upgrade it!

raulk (approved):

PR #276, opened 2020-11-12
scale measurements; fix tags

nonsense:
This PR is fixing the sending of metrics from `graphsync` to:
1. Scale metrics to seconds range, since the results we get are between 1 and 80sec.
2. Add unit (`sec`) to the measurement so that it is clear that we are emitting seconds.
3. Add humanized tags for `bw` and `latency`.
4. Using the `go-sdk` from `master`, since it contains the update storing of metrics (i.e. including the testplan and testcase in the measurement name).

### DIFF
--- a/graphsync/go.mod
+++ b/graphsync/go.mod
@@ -3,6 +3,7 @@ module github.com/libp2p/test-plans/ping
 go 1.14
 
 require (
+	github.com/dustin/go-humanize v1.0.0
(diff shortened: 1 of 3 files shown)


nonsense:
Not much changing in this PR. Merging this as part of the iteration to bring metadata to test plan metrics and the SDK, so that we have real data to review tomorrow.

